Issue #71, opened 2014-07-30
Access global variables defined by included C libraries


HalosGhost:
Continuing my project of messing around with NCurses under Terra, I've discovered that all the functions and type definitions are easily accessible. E.g.:

    C = {}; C.nc = terralib.includec("ncurses.h")
    var mainwin: &C.nc.WINDOW = C.nc.initscr()

The above works just fine. However, when I test to see what all is available in ``C.nc``, it does not appear that variables themselves are made available (assume I have correctly imported ``stdio.h`` as ``C.io``):

    for k, v in pairs(C.nc) do
        C.io.printf("%s ", k)
    end
    C.io.printf("\n")

The above code yields the following list:

[...]

PR #100, opened 2015-06-25
Add support for importing external global variables from C

0----0:
I wanted to interface better with the GLEW library like mentioned in #71, so I did my best to implement the importing of global variables.  It seems to work for both simple cases like
```lua
C = terralib.includec("stdio.h")
terra main()
        C.fprintf(C.stdio, "Hello, world!")
end
main()
```
as well as with more complicated cases, like global function pointers.  (the KeepVarLive seems to be only necessary for function pointers, for some reason.)  If there's anything more complicated that I'm missing, I'm not aware of it.

I've never worked with the LLVM or Lua APIs before, so if you have the time, I'd appreciate it if you could look over my changes.  Thanks.

### DIFF
--- a/src/tcompiler.cpp
+++ b/src/tcompiler.cpp
@@ -1105,7 +1105,18 @@ static GlobalVariable * GetGlobalVariable(terra_State * T, Obj * global, const c
         Obj t;
         global->obj("type",&t);
         Type * typ = Types(T).Get(&t)->type;
-        
+
+        if(global->boolean("isextern")) {
+            const char * _name = name;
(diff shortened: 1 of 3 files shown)


zdevito:
Cool -- thanks for the patch. I'll try to look it over soon. Any time there are changes that use Clang and LLVM APIs it takes a lot of trial and error to see if there are corner cases, but the code generally looks correct.

zdevito:
This should work now. I moved a few things around, so let me know if you run into any problems with your use case.

PR #103, opened 2015-07-11
Return an exit code from the test runner on fail

jrk:
Useful for automated testing (a `test` method in a Homebrew package in my case).


zdevito:
Merged as ddab90cd19d60c0dc204eebf01d0374248b64429

PR #122, opened 2015-09-19
Debian multiarch compatibility

rdebath:
Two commits here, the first is the multiarch updates you requested.
Unfortunately I ran into a bit of a problem, it seems that the clang-dev and llvm-dev packages aren't actually multarch aware yet (for "jessie").  Nevertheless, as all except the final link works correctly I expect it will happen quickly after the packager decides that it needs doing "now". 

The result is that it _should_ compile using "make TARGET_ARCH=-m64" as soon as the libraries become available, but right now the only completely tested part of these changes is actually running the 64bit executable on a, mostly, 32bit system. (which now works correctly for all your tests)

The second commit is a minor tidy-up of the .gitignore files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
-build/
+*.o
+*.a
+*.so
+terra
--- a/Makefile
+++ b/Makefile
@@ -24,9 +24,9 @@ CLANG ?= $(CLANG_PREFIX)/bin/clang
 
 #path to the compiler you want to use to compile libterra
 #can be any c/c++ compiler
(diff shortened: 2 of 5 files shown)


elliottslaughter:
This has been sitting 10 years, and in that time we've rewritten the build system to use CMake, LLVM has been updated ~20 times and a bunch of other things have changed.

I'd imagine you've moved on, but on the very narrow off chance that you (or someone else) finds this valuable, it would need to be updated against the modern architecture.

Closing for now, feel free to submit new issues/PRs for anything else.

PR #139, opened 2015-12-11
clean up a bunch compiler warnings

streichler:
Builds with -Werror on both Linux (gcc 4.8.5, clang 3.5) and OSX (clang 3.5.2).


streichler:
I appear to have be overzealous with the .gitignore change - fix coming shortly.

PR #153, opened 2016-01-23
Conda+Luarocks installable, requirable

alexbw:


terra-scm-1.rockspec
```diff
@@ -0,0 +1,34 @@
...
+   homepage = "",
+   license = "MIT",
+}
+
+build = {
+   type = "command",
+   build_command = [[
+   make clean
+
+    make LUAJIT_LIB=$(LUA_LIBDIR)/libluajit-5.1.a \
```
alexbw: This is the iffy part. It'd be great if we weren't so dependent on very specific library and directory names. Do you think the Makefile is the right place to add robustness?

zdevito: I updated Terra so that changing the luajit used just requires setting the prefix, e.g.

    LUAJIT_PREFIX=/usr/local

This just needs a small modification to match, but I don't have a luarocks install to test it with.

alexbw: Great, so I use `make PREFIX=$PREFIX install` now, works nicely.

LUAJIT_PREFIX doesn't work for me because I set lua headers to install in `$LUAJIT_PREFIX/include`, and not `$LUAJIT_PREFIX/include/luajit-2.0/`. This is to make LuaJIT install identically to PUC Lua. 

Would be good to not hard-code the specific libs, but instead grep for a pattern, and then use the directory in which the desired file resides. This would prepare us for PUC Lua support.

zdevito: I updated the Makefile to search in prefix/include if
include/luajit-2.0/lua.h does not exist, which should fix your issue.

alexbw:
Awesome, thanks for your help, the rockspec is now pretty minimal. 
The rockspec assumes LuaJIT and clang are installed in the same spot, without really stating so, but I don't think there's a way to have luarocks require those types of things, so this might be as good as it's going to get.

alexbw:
And was able to successfully use Terra as a required package in LuaJIT for some little toy problems.

zdevito:
I merged this into master, so just start another pull request if there are additional changes needed.

PR #157, opened 2016-03-14
several fixes for nontemporal loads and stores

streichler:
Three fixes:
1) Generate !nontemporal metadata for loads as well.  (This is allowed in the LLVM IR, although the x86 code generators seem to ignore it for now.)
2) Do not generate the !nontemporal metadata if the "nontemporal" attribute is false.
3) Allow the attributes for attrload/attrstore to be provided via a variable holding a lua table in addition to the "immediate" constructor syntax.

### DIFF
--- a/src/tcompiler.cpp
+++ b/src/tcompiler.cpp
@@ -2055,6 +2055,14 @@ if(baseT->isIntegerTy()) { \
                     int alignment = attr.number("alignment");
                     l->setAlignment(alignment);
                 }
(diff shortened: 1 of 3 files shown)


0joshuaolson1:
@streichler ?

streichler:
Dusted off this pull request.  Item 2 in the list above had already been addressed in master, but the other two fixes are still important (to me, at least).

PR #161, opened 2016-03-19
Add failing hex formatted number test.

brandonbloom:
Fails with:
tests/hexf.t:18: malformed number near '0xfffffffffffffffff'

### DIFF
--- a/tests/hexf.t
+++ b/tests/hexf.t
@@ -7,3 +7,15 @@ terra foo()
 end
 
 assert(foo() == 15)
+
+terra u32() : uint32
+	return 0xffffffff
+end
+
+assert(u32() == math.pow(2, 32) - 1)
+
+terra u64() : uin64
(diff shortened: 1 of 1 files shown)


brandonbloom:
The 32-bit test passes, but the 64-bit test fails to parse. I took a brief look at how number scanning works, but didn't see any obvious problems. Happy to dig deeper and attempt a fix if it's not a quicky fix for you.

zdevito:
Similar to C, number literals are assumed to be 32-bit by default. If you
want a 64-bit unsigned integer literal you can write it like:
 terra u64() : uint64 return 0xffffffffffffffffULL end

brandonbloom:
Ah, OK - I see that in the docs now. Thanks.

I've been writing too much Go, where the literals are polymorphic: https://golang.org/ref/spec#Constants

PR #170, opened 2016-05-07
add entrymissing must be macro

sarostru:
I was just trying out the entrymissing capability for the first time and at first I tried using a lua function rather than a macro, reading the terra source showed me that it had to be a macro, but maybe worth having in the api reference.


zdevito:
Sure, thanks.

Issue #126, opened 2015-10-23
README's assumptions about Terra binary location


0joshuaolson1:
For the Linux release at least,
`$ ./terra tests/hello.t`
should be
`$ bin/terra share/terra/tests/hello.t`
But the file hierarchy is different when building from source, is it not? I think correcting this would either require complicating the readme or using symlinks (which don't work on Windows?).

zdevito:
Yes, they can be updated to refer to the `release` directory. It should
have the same structure whether in the source code or binary distribution
(in the source, the binary directory is just <repository>/release).

0joshuaolson1:
I admit that I don't understand what you mean. I haven't built from source, so I'm guessing the other version would look like `build/SOMETHING/terra tests/hello.t`?

elliottslaughter:
I think Terra's Makefile has installed a symlink from `release/bin/terra` to `terra` for a long time now.

On the other hand, CMake doesn't work this way, but it's really impossible to know what path to put since the user could install it anywhere. I don't think it's unreasonable to assume the user will know where they installed Terra?

At any rate, I'm closing this unless someone wants to take a look at the modern READE file and give a concrete suggestion for what to change. (As per the original issue title, PR's are great too.)

PR #185, opened 2016-06-30
Update getting-started to match readme

0joshuaolson1:
Except for bin/terra -> ./terra and share/terra/tests -> tests as per [#126](https://github.com/zdevito/terra/issues/126)

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -7,6 +7,15 @@ title: Getting Started
 * auto-gen TOC:
 {:toc}
 
+Getting Started with Terra
+==========================
+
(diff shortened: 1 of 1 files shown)


zdevito:
I think in this case the getting-started guide is more up-to-date than the readme, so the patch should probably go in the other direction except for the typos that were fixed in the readme.

0joshuaolson1:
I'll try again.

0joshuaolson1:
Too fast! I didn't get to say that [#186](https://github.com/zdevito/terra/pull/186) is only half of this - I'll fix this PR too.

0joshuaolson1:
I fixed 'download the download the' and added the first section that was only in getting started.

0joshuaolson1:
This is ready now.

PR #206, opened 2016-08-18
Modified golike.t to work with terralib.saveobj

Lupus:
This PR introduces modified golike.t, which now uses constant expressions for vtables. This has two benefits --- (1) golike.t based code can now be saved to object file and (2) more room for compiler optimizations.


elliottslaughter:
It's been a while, but I figured this should get merged. Thanks!

PR #209, opened 2016-08-24
Extend support for :asvalue() + test

nicolasvasilache:
Added some more cases for asvalue() and a unit test that I find useful in what I am doing (line 871 ..).
About all the line changes, my emacs config automatically removes trailing whitespaces; let me know if this is too annoying and I'll submit another PR without those.

### DIFF
--- a/src/terralib.lua
+++ b/src/terralib.lua
@@ -47,7 +47,7 @@ ident =     escapedident(luaexpression expression) # removed during specializati
 
 field = recfield(ident key, tree value)
       | listfield(tree value)
-      
+
 structbody = structentry(string key, luaexpression type)
            | structlist(structbody* entries)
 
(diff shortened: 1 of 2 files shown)


aiverson:
Constant folding in :asvalue looks very useful.

Does this handle operator overloading on custom types?

Can we integrate this change? Can we fix the whitespace on everything in master?

elliottslaughter:
It would definitely be easier to merge without the whitespace changes. If @nicolasvasilache or someone else has time to do that it would be great, otherwise I may get to it eventually (but don't wait on me, eventually might be a long time).

aiverson:
I've found a few bad cases that should be mentioned related to this. For one, (`nil):asvalue() is difficult to tell apart from a failed constant resolution. Also, it currently doesn't resolve things like (quote in 5 end):asvalue()

I think adding an isvalue/isconstant method and an istype method to check for a quote being a constant value or a type would relieve the difficulty of checking for errors and remove the ambiguity/failure case where the value of the quote may be nil.

aiverson:
I'd like to request that the constant folding additions included here also be applied to the isconstant function. Or some way to test whether something is a constant or not, and get the value. Currently, quote:asvalue will return a symbol when given a quote of looking up the value at a variable, which doesn't fit the usecase of extracting a constant value from a quote.

aiverson:
As a corollary to my last comment, I'd like to propose that any operation needed by the standard library that needs any awareness of the internal representation of trees be made part of the public API of terralib with respect to quotes so that refactors affecting the AST like the addition of new node kinds or anything can remain more interface compatible with less breakage.

I'm not sure that this is entirely possible to do, but it seems like a good goal.

PR #224, opened 2016-11-30
better luarocks

andresy:


andresy:
you might want:
(1) to document it is important to provide CLANG and LLVM_CONFIG
(2) check that CLANG and LLVM_CONFIG are actually viable
(3) warn that this works only if the luajit lib is called libluajit.{so,dylib} (unfortunately, luarocks does not always ship LUALIB)

PR #232, opened 2017-03-20
Fixed max implementation in terra for c++ examples on website

Mx7f:


elliottslaughter:
It's been a while, but I figured this should be merged. Thanks!

PR #244, opened 2017-08-14
MakeFile Included file for Mac

roachmd:
Might be Helpful.

Using homebrew
brew install llvm@3.7

Rename Mac_Makefile.inc to Makefile.inc


iology:
Hard-coding the paths is definitely not a good idea. Also, `llvm-config --ldflags` and `llvm-config --cxxflags` give the necessary settings. Does the option `-nostdinc++` you wrote really affect the building process? (with regard to Homebrew, MacPorts, ..., and manual build process) Maybe you can instead provide a brew formular.

PR #247, opened 2017-09-11
Fix for non-BSD Sed

elliottslaughter:
The `-E` flag for sed is BSD-specific. It is supported for backwards compatibility only in certain newer versions of GNU Sed, but not all versions. (Furthermore, this support is [undocumented](https://stackoverflow.com/a/3139925/188046).)

This patch checks that `sed -E` actually works, and otherwise falls back to `sed -r` which is the GNU Sed version of this flag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,14 @@ LD = ld
 FLAGS += -Wall -g -fPIC
 LFLAGS = -g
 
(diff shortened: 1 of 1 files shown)


elliottslaughter:
@zdevito Ping.

PR #249, opened 2017-10-13
build: optionally re-export some/all LLVM components

streichler:
This change makes it easier to use LLVM-enabled third-party libraries (e.g. Legion) in terra applications.
Starting with LLVM 3.6+, it's a fatal error to have two copies of some of the LLVM objects in the same process, but it's hard to have a third-party library use the LLVM code that has been imported into terra
because the terra build includes only the portions of the libraries it needs.

With this change, the default build remains the same, but if the REEXPORT_LLVM_COMPONENTS
variable is set by the caller, any libraries needed for that component (i.e. the result of 
`llvm-config --libs ${REEXPORT_LLVM_COMPONENTS}`) are included in the terra binary in their entirety.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,22 @@ ifneq (,$(findstring $(LLVM_VERSION),$(CLANG_REWRITE_CORE)))
 LLVM_LIBRARY_FLAGS += -lclangRewriteCore
 endif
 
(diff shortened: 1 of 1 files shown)


elliottslaughter:
@streichler Please rebase master so that we can see if tests pass. Thanks.

PR #260, opened 2017-12-14
Add more informative error messages in linkllvm()

Mx7f:
That might save someone some time if, to take a definitely hypothetical example, they performed some tensorflow installation on windows that installed a different version of CUDA and as part of the install process, renamed the CUDA_PATH environment variable to specify the version, thus breaking CUDA support, manifesting in a large terra codebase that uses linkllvm as a cryptic "llvm: no such file or directory" error that then took an embarrassing amount of time to track down.


elliottslaughter:
This looks fine to me. I'll merge in just a sec.

elliottslaughter:
Github doesn't see it as merged, but it's merged.

PR #262, opened 2017-12-16
Update develop branch

Clyybber:
Side question: Should future pull request made to develop or master?


elliottslaughter:
What is this pull request about?

My understanding, though @zdevito should correct me if I've got this wrong, is that we were planning on cherry-picking anything important from develop into master, and leaving the rest. (There were concerns with whether the current `develop` changes would be fully backwards compatible or not.) So any new work that you want to see merged should also target master.

elliottslaughter:
Sorry, I see now this is doing `master -> develop` not `develop -> master`. That answers my first question. I think the rest of my previous comment still stands.

elliottslaughter:
I should add: if there are specific changes in develop that you care about, please identify those and we can work on cherry-picking them into master.

Clyybber:
Ok, should this still be merged? I guess it's easier to maintain if develop has the same base as master.

Clyybber:
ping @elliottslaughter

elliottslaughter:
My inclination would be to make this a rebase, and to specifically exclude anything that has already moved into master (i.e. Travis support). That way the set of commits on develop should shrink over time and it's obvious what is still outstanding (which can become hard to see when you merge). I'm guessing no one actually uses develop directly, so if we mess with history that's hopefully not a problem.

Clyybber:
As @elliottslaughter said a rebase would be more appropriate.

PR #265, opened 2018-01-25
attempt at adding support for FreeBSD

streichler:
The first change allows terra to build and run, at least in the FreeBSD virtual machine I'm using.
I built with clang 3.8, and gmake is required (the BSD version of make does not like the Makefiles at all).
There is no CUDA for FreeBSD, so this is CPU only.  One test (stattest.t) is disabled because FreeBSD's
sys/stat.h uses bit fields in `struct stat`, which confuses terra.

The second two changes address an issue that may or may not be specific to my VM.  Despite the OS
claiming that the CPU supports AVX instructions, I get a SIGILL in many tests because llvm happily
generates AVX instructions.  The workaround involves a general-purpose change that is needed to actually get feature flags into MCJIT and clang (for include_c), followed by a ifdef-protected change
that makes "-avx" the default feature flag if `-DDISABLE_AVX` is set via the `FLAGS` make variable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 /build
 /release/bin
 /release/lib
+/release/include/terra/lauxlib.h
+/release/include/terra/lua*.h
 /terra
 /tests/afile.txt
 /tests/benchmark_fannkuchredux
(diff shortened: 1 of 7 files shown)


streichler:
CI identified a couple bugs with my change on non-FreeBSD systems, which I believe I have cleaned up.  I also reordered the changes so that the generally-useful ones (now three, including an update to the gitignore file) come before the FreeBSD-specific one.

PR #270, opened 2018-02-24
:art: :new: Command line option -e

jleija:
Problem:
- terra does not support command line option -e
- this makes it difficult to use terra as a drop-in replacement for other lua implementations when doing scripting
- it is useful to evaluate/run a terra chunk of code in the command line

Solution:
- add a new command-line option -e with the same semantics as other lua implementations

Example: ./terra -e 'c = terralib.includec("stdio.h"); c.printf("Hello World!\n")'

### DIFF
--- a/release/include/terra/terra.h
+++ b/release/include/terra/terra.h
@@ -18,6 +18,7 @@ typedef struct {   /* default values are 0 */
     int verbose; /*-v, print more debugging info (can be 1 for some, 2 for more) */
     int debug;   /*-g, turn on debugging symbols and base pointers */
     int usemcjit;
+    char* cmd_line_chunk;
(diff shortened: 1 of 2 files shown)


elliottslaughter:
This looks ok to me. The change is pretty simple, but I'm wondering if we can get some more community involvement in looking at patches. @Clyybber, would you be interested in doing a code review? I think it shouldn't take too long, the code is pretty short. If that goes well we could make this a more regular part of the development process. (If we can't find someone I'll just merge it.)

Clyybber:
Yeah, thank you, I'll go ahead and review it.

elliottslaughter:
Thank you both!

Issue #273, opened 2018-04-10
 LLVM can't generate the entry of local function in the function:"JITGlobalValue"


wenjieshen:
Hi all,

The version of my LLVM is 3.8.0.
Terra occurred a problem that I can't generate completed same local function for second JIT module.
In the test(class6.t), the Terra should generate two JIT module but Terra just generate :

`declare internal double @"$C.times2"(%C*)`

instead of

```
declare internal double @"$C.times2"(%C*) {
entry:
    %1 = getelementptr %C, %C* %0, i64 0, i32 1
    %2 = load i32, i32*%1, align 4
    %3 = shl i32 %2, 2
    %4 = sitofp i32 %3 to double
    ret double %4
}

```
LLVM will crash when it can't link the function.

Thank you for your attention.

elliottslaughter:
The linked PR has merged, so I am closing this issue. Thanks again for the fix!

PR #274, opened 2018-04-17
Update tllvmutil.cpp to solve issue

wenjieshen:
Set linkage to external linkage when the function is existed in module #273

### DIFF
--- a/src/tllvmutil.cpp
+++ b/src/tllvmutil.cpp
@@ -245,6 +245,8 @@ struct CopyConnectedComponent : public ValueMaterializer {
                 VMap[fn] = newfn;
                 SmallVector<ReturnInst*,8> Returns;
                 CloneFunctionInto(newfn, fn, VMap, true, Returns, "", NULL, NULL, this);
+            }else{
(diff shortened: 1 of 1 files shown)


elliottslaughter:
Is it possible to add a test case for this? (Ideally one that fails without this patch and passes with it.)

Thanks!

wenjieshen:
Unfortunately, It only occurred on specific platform(windows 10,LLVM3.8.0 and built by MSVC) and that already failed on existing tests( class6.t, coverage3.t...etc).

elliottslaughter:
Thanks, looks good.

I don't know if you're interested, but it would be really nice if you or someone had time to set up appveyor for the project. At the moment I don't have a usable Windows dev box and so I can't really respond to a lot of Windows-specific issues that come up, or even offer advice on a setup that is known to work. If we automated tested we'd at least have a known working build recipe. (And even if it didn't pass all tests, it would still be a big step forward.) Anyway, if you have any interest in this, let me know or feel free to send another pull request.

PR #278, opened 2018-05-15
Handle libdevice properly on CUDA 9.0+

manopapad:
When `saveobj`-ing a CUDA module, Terra will now automatically find libdevice (an LLVM bitcode library that contains math functions for GPUs) and link it in. Because we now set the correct target for CUDA modules, LLVM knows to run the appropriate passes (in particular NVVMReflect, which will desugar the `__nvvm_reflect` pseudo-instruction, which is used in libdevice for CUDA versions 9.0 and up).

The user (still) needs to manually `linkllvm` libdevice, if they want to run CUDA code on the terra process.

### DIFF
--- a/src/cudalib.lua
+++ b/src/cudalib.lua
@@ -28,7 +28,7 @@ terralib.CUDAParams.entries = { { "gridDimX", uint },
                                 
 function cudalib.toptx(module,dumpmodule,version)
     dumpmodule,version = not not dumpmodule,assert(tonumber(version))
-    local cu = terralib.newcompilationunit(terralib.nativetarget,false) -- TODO: add nvptx target options here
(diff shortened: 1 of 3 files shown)


chichengumich:
Hi, I was trying to compile terra with CUDA 9.0 and llvm 3.8. The compliation went through without any error. However, when I tried to run cudahello.t, it returns
```
cudahello.t:21: cuModuleGetFunction: cuda reported error 500
stack traceback:
	[C]: in function 'error'
	[string "<string>"]:247: in function 'cudacompile'
	cudahello.t:21: in main chunk
```
Is there any solution?

elliottslaughter:
@chichengumich Please reply to this issue with information to help us reproduce your bug: https://github.com/zdevito/terra/issues/288 . Thanks.

Issue #295, opened 2018-08-01
Construction of llvm_list file fails during build


lightsighter:
This line of the Terra build Makefile doesn't work properly with Ubuntu 16.04, GNU Make 4.1, and GCC 5.4.0:

https://github.com/zdevito/terra/blob/master/Makefile#L207

Instead of writing the results to the expected file, it dumps all the file names to the terminal screen. The build then fails later because the llvm_list file is empty.

elliottslaughter:
Fixed by #300.

Issue #235, opened 2017-04-18
Terra doesn't work with LLVM compiled with CMake


elliottslaughter:
I'm seeing some bizarre errors that occur only when LLVM has been compiled with CMake, but which go away when I build with autoconf. Here's an example of what I see when I run `make test`:

```
terra: .../Support/Casting.h:237: typename llvm::cast_retty<X, Y*>::ret_type llvm::cast(Y*)
  [with X = clang::ObjCObjectPointerType; Y = const clang::Type;
   typename llvm::cast_retty<X, Y*>::ret_type = const clang::ObjCObjectPointerType*]:
  Assertion `isa<X>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
```

I've done the best I can to make sure the two builds are as similar as possible, though obviously there has to be some difference between them. Here are the two sets of flags:

```
../llvm-3.8.1.src/configure --prefix=$PWD/../install \
  --enable-optimized \
  --disable-assertions \
  --disable-terminfo \
  --disable-libedit \
  --disable-zlib
make
make install

cmake -DCMAKE_INSTALL_PREFIX=../install_cmake \
  -DCMAKE_BUILD_TYPE=MinSizeRel \
  -DLLVM_ENABLE_ASSERTIONS=OFF \
  -DLLVM_ENABLE_ZLIB=OFF \
  ../llvm-3.8.1.src
make
make install
```

MinSizeRel (rather than Release) is necessary because CMake tries to use GCC's LTO, which generates massive object files (total > 25 GB) that then fail to link later in the Terra build process. That means CMake is using `-Os` rather than `-O3` for the binaries; but I would be surprised if this could be the issue. I can't find flags for CMake to disable terminfo and libedit, though again I sincerely doubt those are causing the issue. I can't see anything else that would make a difference. Assertions are disabled in both builds.

I'm on Ubuntu 16.04 and using GCC 5.4 to build LLVM. In fact I started looking into this because I couldn't get the Ubuntu packages to work and I suspect they've moved to CMake in their build process as well.

elliottslaughter:
Saw this again recently while testing the new CMake build system #297. Different backtrace, but otherwise similar symptoms (occurs only with CMake, not autoconf). LLVM 3.8, Ubuntu 16.04, GCC 5.5.0, CMake 3.5.1.

[...]

PR #297, opened 2018-08-17
Add support for CMake-based build system

elliottslaughter:
This pull request adds a new build system based on CMake. The main advantage of CMake is that since LLVM has switched over to it, those libraries can now be found automatically. The old Makefile-based build system used a number of ugly workarounds to do the same thing, leading to bugs like [this one](https://github.com/zdevito/terra/issues/295) that are essentially impossible to fix in the existing infrastructure.

I have no plans to remove the Makefile based build, but going forward I suggest that CMake will be much easier to maintain and should be the focus of our efforts.

Currently this PR is mainly here to make sure the build gets tested. There are a number of missing features that will need to be added before this is ready for all the use cases currently supported by the existing Makefile.

Caveats:

  * Requires LLVM >= 3.6. It does not matter if LLVM was built with autoconf or CMake because the autoconf build also spits out the files CMake needs to find everything.
  * [DONE] Currently this only works on Linux. Will test macOS next.
  * [DONE] Not yet tested with CUDA.
  * Definitely not compatible with Windows. This may be an eventual goal but will require more work.

Other things I'd like to do at some point if I can find the time:

  * [DONE] Make the build fully out-of-tree. Right now there are a bunch of hacks that basically need it to be built in a very specific way. But changing this will require changing the test infrastructure too.
  * [DONE] Support for a proper installation option. The current one is kind of hack that only copies files into the `release` directory.


elliottslaughter:
Soliciting feedback from any potentially interested parties. Comments welcome.

@OvermindDL1 @Clyybber @manopapad @lightsighter (and anyone else with an interest in this)

Clyybber:
This is great, much more cleaner than the Makefile.

OvermindDL1:
I'd might also recommend using [hunter.sh](https://docs.hunter.sh/) or so to pull [LLVM](https://docs.hunter.sh/en/latest/packages/pkg/LLVM.html) directly.  This way you can specify the precise version you want terra to use with the precise build options and all.  Hunter.sh itself is just a 'HunterGate.cmake' file you include in the project and you specify a 'package repository listing' (like linked above via LLVM is the generic 'standard' one, but it is trivial to make up your own as a git repo or file download or so), it then compiles it with the options needed in a global area for re-use (overrideable of course) and exposes it into the application for full normal Modern CMake target linking.  This makes the builds entirely and fully reproducible and lowers the barrier of entry and usage (and people using wrong LLVM versions) without requiring any other dependencies (I.E. since Hunter.sh is a fully cmake dependency manager it will work with any setup and any build system without requiring anything but cmake itself to be installed).

And yes, CPack should definitely be used to build installers (Windows, sh, deb, etc... etc...).

elliottslaughter:
Oh boy, is LLVM ever a mess. It seems like nearly every configuration that I don't compile myself is broken in one way or other.

Ubuntu 16.04:
  * LLVM 3.8 from source: compiles and passes all tests.
  * LLVM 3.8 from Ubuntu package repo: broken because the packagers decided to rename `/usr/share/llvm` to /usr/share/llvm-3.8` and forgot to update all the paths to point to the new location.
  * LLVM 3.8 from llvm.org binaries: broken with link errors due to what looks like a mismatch in C++11 symbols, but the usual tricks (`-fabi-version=2 -D_GLIBCXX_USE_CXX11_ABI=0`) don't fix it.
  * LLVM 3.8 from apt.llvm.org packages for Ubuntu: I somehow got my machine into a state where I can't install these, but I believe they compiled but segfault when running.

Ubuntu 14.04:
  * LLVM 3.8 from source: compiles and passes all tests.
  * LLVM 3.8 from Ubuntu package repo: broken, as above.
  * LLVM 3.8 from llvm.org binaries: compiles, but then segfaults when running.
  * LLVM 3.8 from apt.llvm.org packages for Ubuntu: compiles, but then segfaults when running.

Ouch.

I looked into Hunter a bit, and I like the basic idea (and especially how it's mostly invisible to end users), though I'm a bit concerned that as with any small-ish packaging system, the size of the community is going to be problematic. (This is my biggest reason for using Homebrew over MacPorts on macOS: packages are just that much more likely to be up-to-date and maintained when there are that many more people using the system.) However, practically speaking, the biggest issue is that they currently only seem to support [LLVM 3.6 and 4.0](https://github.com/hunter-packages/llvm), and the way packages are managed seems to involve unpacking the entire source tree into a Git repo and then modified by hand (!?!) so it's not super friendly to people doing it themselves.

OvermindDL1:
> though I'm a bit concerned that as with any small-ish packaging system, the size of the community is going to be problematic. 

Huntergate is just the manager for it all.  The 'repository' used as an example by docs does have a lot of packages, however it is designed not to be centralized, you can fork it, can make your own from scratch, can use multiple repositories all at once, etc...  It was designed to be able to persist without any central location, which goes well with its 'invisible to the end user' aspect.

However I do not see it failing in any way, it has a clear set of goals and it is managed by one of the big cmake developers so it would not be unexpected to see huntergate be distributed with cmake itself one day as well considering how comparatively simple it is.

> Oh boy, is LLVM ever a mess. It seems like nearly every configuration that I don't compile myself is broken in one way or other.

But yes, this right here is why I will *not* touch LLVM from a package repository.  LLVM was not even designed to be used from a package repository as such considering how much of it is supposed to be configured at its own build time, and that's even before considering that they always seem to be highly out of date...

elliottslaughter:
@OvermindDL1 Would you be willing to look into getting LLVM 3.8 included in the hunter.sh packages repo? Then I could at least play around with it.

For what it's worth, I think my CMake bug is not an instance of #235. The 14.04/16.04 packages are still definitely broken, but at least for 18.04 and built from source with CMake it should be possible to get CMake working.... I just have to figure out what's still different between the two builds.

OvermindDL1:
> @OvermindDL1 Would you be willing to look into getting LLVM 3.8 included in the hunter.sh packages repo? Then I could at least play around with it.

Yeah it looks like no one has needed 3.8 so it was never added:
https://github.com/ruslo/hunter/blob/7f0c446acaf17d1ab91c8150cf3065947d91db15/cmake/projects/LLVM/hunter.cmake
It currently has:  3.4.2 3.5.0 3.6.0 3.6.2 3.6.2-p0 4.0.1-p0

They are generally quite open about PR's and/or requests, I'll go ahead and submit a request for both 3.8 as well as the latest 5.x and the in-dev 6.x.

Feature request submitted:  https://github.com/ruslo/hunter/issues/1522

fabiencastan:
Is there still a plan to integrate a cmake build system in terra?

elliottslaughter:
Yes, this is still the plan, I just haven't had time to get back to it yet.

elliottslaughter:
Ok, the CMake build is now passing tests locally, so at this point I think it's just a matter of making Travis happy again.

The one thing I'm unhappy about at this point is that I couldn't get the tests to pass without extracting all the `.o` files from the static libraries and linking them in individually. I honestly don't understand how this is any different than `-Wl,--whole-archive`, but I can't get the tests to pass that way, so I guess we're stuck with this for now. It does mean that we rely on `CMAKE_AR` being set, so I'm not sure this will be portable.

One option would be to add a flag to specify dynamic linking, but you'd never get all the tests to pass under such a configuration.

elliottslaughter:
I'm taking this pull request since all the basic features work, and there are systems where you basically can't build without it. Mac/Linux have been reasonable well-tested, including with CUDA. I'm leaving Windows and Hunter support to be revisited in the future. I've done some initial work to try to avoid making this platform-specific, but there will certainly be more to do when I try to go for full Windows support.

I'd still appreciate testing from anyone in this thread who is interested in CMake support. You can get it directly from master now that the PR has merged.

PR #300, opened 2018-08-24
Fix order of Terra redirects causing output not to get captured on stderr

elliottslaughter:
Potential fix for #295.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,7 +209,7 @@ release/include/terra/%.h:  $(LUAJIT_INCLUDE)/%.h $(LUAJIT_LIB)
     
 build/llvm_objects/llvm_list:    $(addprefix build/, $(LIBOBJS) $(EXEOBJS))
 	mkdir -p build/llvm_objects/luajit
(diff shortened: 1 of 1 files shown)


lightsighter:
No the patch still doesn't fix it. I checked out a fresh copy of terra and manually applied the patch before trying to build. I still see all the file names getting dumped to the screen instead of going into the llvm_list file. The problem should be with the Makefile rule on line 210 which builds 'llvm_list'.

elliottslaughter:
@lightsighter Can you reclone and try again with the latest patch?

elliottslaughter:
A simple test that confirms that the relative ordering of `2>&1` matters:

```
$ cat test_stderr.sh 
echo stdout
echo stderr 1>&2
$ chmod +x test_stderr.sh
$ ./test_stderr.sh > /dev/null
stderr
$ ./test_stderr.sh 2> /dev/null
stdout
$ ./test_stderr.sh 2>&1 > /dev/null
stderr
$ ./test_stderr.sh > /dev/null 2>&1
```

elliottslaughter:
@lightsighter Confirmed the fix, taking this patch.

PR #302, opened 2018-08-25
Add limits to numeric types

magnatelee:
This patch provides a reliable way to get the min and max values of each numeric type.


src/terralib.lua
```diff
@@ -1651,28 +1651,38 @@ do
...
-            local name = "int"..tostring(size * 8)
+            local bits = size * 8
+            local name = "int"..tostring(bits)
             if not s then
                 name = "u"..name
             end
+            local min,max
+            if not s then
+                min = 0ULL
+                max = 2ULL ^ (bits - 1)
```
elliottslaughter: Shouldn't this be `(2ULL ^ bits) - 1`? Technically this does wrap about at 64 bits, but that seems to work on LuaJIT 2.0.5.

magnatelee: Yup, that was incorrect. Commit 0efbb39 should address this.

elliottslaughter:
Thanks!

PR #310, opened 2018-09-21
LLVM 7.0 Support with CUDA

ProfFan:
LLVM 7.0 is not a huge deal however there is one crucial breaking change: The LLVM repo now starts to have C++ files with same filenames. same source name -> same object name, and when compressed into one `.a` archive, the second one never gets uncompressed, leading to a link error.

Sadly, no elegant cross-platform solution is known to me that handles this situation well. Either we report this to the upstream, or we use some other program to uncompress the archive.

My solution is simple but not elegant: I added a hard dependency to `p7zip`, which is the only thing known to me that works with the 1970-era ar format.

Travis scripts are already adjusted to accommodate this change.

The code explains for itself, BTW.

### DIFF
--- a/src/tllvmutil.cpp
+++ b/src/tllvmutil.cpp
@@ -191,9 +191,16 @@ bool llvmutil_emitobjfile(Module * Mod, TargetMachine * TM, bool outputobjectfil
     #else
     emitobjfile_t & destf = dest;
     #endif
+
+    #if LLVM_VERSION >= 70
+    if (TM->addPassesToEmitFile(pass, destf, nullptr, ft)) {
+        return true;
+    }
+    #else
     if (TM->addPassesToEmitFile(pass, destf, ft)) {
         return true;
     }
+    #endif
 
     pass.run(*Mod);
     
@@ -403,7 +410,18 @@ error_code llvmutil_createtemporaryfile(const Twine &Prefix, StringRef Suffix, S
 int llvmutil_executeandwait(LLVM_PATH_TYPE program, const char ** args, std::string * err) {
 #if LLVM_VERSION >= 34
     bool executionFailed = false;
+    #if LLVM_VERSION >= 70
+    std::vector<llvm::StringRef> argsRef;
+    int args_ptr = 0;
+    while (args[args_ptr] != NULL) {
+        argsRef.push_back(llvm::StringRef(args[args_ptr]));
+        args_ptr++;
+    }
+
+    llvm::sys::ProcessInfo Info = llvm::sys::ExecuteNoWait(program, argsRef, llvm::Optional<ArrayRef<llvm::StringRef>>(), {}, 0, err, &executionFailed);
+    #else
     llvm::sys::ProcessInfo Info = llvm::sys::ExecuteNoWait(program,args,nullptr,{},0,err,&executionFailed);
+    #endif
     if(executionFailed)
         return -1;
     #ifndef _WIN32
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
   - LLVM_CONFIG=llvm-config-3.5 CLANG=clang-3.5
   - LLVM_CONFIG=llvm-config-5.0 CLANG=clang-5.0
   - LLVM_CONFIG=llvm-config-6.0 CLANG=clang-6.0
+  - LLVM_CONFIG=llvm-config-7.0 CLANG=clang-7.0
 
 matrix:
   exclude:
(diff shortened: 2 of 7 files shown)


ProfFan:
Plus that I am seeing possible performance regressions with CUDA on my machine with macOS. Please test with other configurations as well. I think LLVM 7 probably changed some optimizations to non-default. No sure though.

ProfFan:
Please also see https://travis-ci.org/zdevito/terra/jobs/431465551

There is a reproducible crash here. Possibly due to excessive memory usage, but not sure.

Maybe I just linked in some non-required llvm objects?

elliottslaughter:
Thanks again @ProfFan for doing this.

I've been thinking about `.o` conflict issue. This is really only an issue for users who:

  * Use Terra as a library (instead of running the executable)
  * Want to statically link Terra
  * Want Terra to transitively statically link its dependencies
  * Want the Terra static library to also include LLVM and LuaJIT instead of having the user be responsible for linking those separately

In particular, it's *not* an issue if you use the Terra executable or dynamic library, because we can just use `--whole-archive` on those to include LLVM/LuaJIT. It's also not an issue if you're willing to let Terra dynamically link its dependencies.

With the new CMake-based infrastructure, I'm thinking that it should be possible to give people better options. I can add flags to allow dynamic linking of dependencies, or to allow Terra to exclude LLVM/LuaJIT from its static library. With these options in place, we can serve (what I suspect to be) the vast majority of users, and those who really care about the use case above can stick with LLVM 6 or below.

src/tllvmutil.cpp
```diff
@@ -403,7 +410,18 @@ error_code llvmutil_createtemporaryfile(const Twine &Prefix, StringRef Suffix, S
 int llvmutil_executeandwait(LLVM_PATH_TYPE program, const char ** args, std::string * err) {
 #if LLVM_VERSION >= 34
     bool executionFailed = false;
+    #if LLVM_VERSION >= 70
+    std::vector<llvm::StringRef> argsRef;
+    int args_ptr = 0;
+    while (args[args_ptr] != NULL) {
+        argsRef.push_back(llvm::StringRef(args[args_ptr]));
```
MaskRay:
```cpp
    llvm::sys::ProcessInfo Info;
#if LLVM_VERSION < 70
    Info = llvm::sys::ExecuteNoWait(program,args,nullptr,{},0,err,&executionFailed);
#else
    Info = llvm::sys::ExecuteNoWait(program, llvm::toStringRefArray(args),
                                    llvm::None, {}, 0, err, &executionFailed);
#endif
```

MaskRay:
Well, llvm 8 will be a bigger deal:

* `clang::vfs` -> `llvm::vfs` change
* `llvm::JITEvent_EmittedFunctionDetails` is deleted

For the first point, here is what I use in https://github.com/MaskRay/ccls/blob/master/src/clang_tu.hh
```cpp
#if LLVM_VERSION_MAJOR < 8
// D52783 Lift VFS from clang to llvm
namespace llvm {
namespace vfs = clang::vfs;
}
#endif
```

MaskRay:
> My solution is simple but not elegant: I added a hard dependency to p7zip, which is the only thing known to me that works with the 1970-era ar format.

What kind of ar files do you want to process? llvm-ar (shipped with releases.llvm.org prebuilt archives) is versatile.

ProfFan:
@MaskRay Thank you for the suggestions! On the API change, I think `llvm::JITEvent_EmittedFunctionDetails` is a piece of dead code, a ghost dwelling in the past, and relevant code are for compatibility of LLVM 3.x. So a simple ifdef should make it work.

On `ar` format, the CMake-based build solution should make the Makefile-based one obsolete. One question is that, can `llvm-ar` handle the situation when you have an `ar` file with two `.o`s with the same name?

MaskRay:
Can you give me some context why ar/p7zip is used to unpack `.a` files? If you just want to link them, as @elliottslaughter said, you can just use `-Wl,--whole-archive -lyour-archive -Wl,--no-whole-archive`.

elliottslaughter:
Historically, Terra has been distributed three ways: as an executable, a static library, and a dynamic library. For the executable and dynamic library, you're right that `--whole-archive` gets the job done. But for the static library, there isn't any notion of "linking" another library. The closest you can get is to include the library by containment: i.e. to literally extract the `.o`s and include them in your own `.a`. This is what Terra has historically done. The other (most standard) option is to do something like a `pkg-config` type approach where you include some way to locate dependencies---and then the user is responsible for linking those whenever they use your library. While I wasn't around at the time this decision was being made, I believe Terra's approach was chosen for usability. LLVM is a big, complicated dependency with multiple components (at least when packaged as a static library), and so it's particularly nice for users not to have to think about that, especially if they're not doing anything else with LLVM aside from use it via Terra. It also makes arranging releases easier, particularly in the absence of package managers, since there are only the three Terra artifacts (executable and static/dynamic libraries) need to be included, not any dependencies.

The reason why this came up in the context of LLVM 7 specifically, is that within LLVM 7 there is a least one `.a` which contains multiple `.o`s with the same name. Presumably, what happened is that these `.o`s live in different directories, so there isn't actually a name conflict in the LLVM build itself. And usually when linking to a `.a`, the names of the `.o` files aren't considered at all, so it's irrelevant. But for this use case it is problematic because apparently `ar` isn't smart enough to apply even a basic conflict resolution algorithm; it'll just silently fail to unpack one of the `.o`s in the archive, and then you get unresolved symbols.

The initial workaround for this---which I think is completely reasonable given the circumstances---was to use the 7-zip archiver. Unlike `ar`, 7-zip is smart enough to rename the files in some sort of a semi-intelligent way on conflicts. We don't actually care about the file names, we just care about not losing the contents, so that is good enough for us. The issue though is that this makes 7-zip a build-time dependency, at least in LLVM 7 (and presumably newer versions).

However, another thing that has changed since this PR was first opened is that we've grown a CMake-based build system, and as of #321 that build system has the ability to work around this issue by using `--whole-archive` instead of extracting the `.o`s from the LLVM libraries. So another option would be to drop the 7-zip dependency and say that, at least with LLVM 7 and up, you have to build with CMake, and you have to set `TERRA_SLIB_INCLUDE_LLVM` to false. This won't break any existing users, because it'll leave the existing build system intact for currently supported set of LLVM version. If this turns out to be problematic for Terra users, we can revisit it in the future. But otherwise, we can begin to slowly move away from this feature which has been so problematic for maintenance.

MaskRay:
OK, if the object files with the same name are not too many, you can use `ar xN 2 libfoo.a a.o` to extract the 2nd one, when targeting GNU ar. When you get the future release of llvm 8, its llvm-ar supports an option `L` to add the contents of an archive to another archive (https://reviews.llvm.org/D53332)..

elliottslaughter:
Just so I understand, is `llvm-ar` something we can find in another (possibly non-default) package prior to LLVM 8, or is that really the first time it will be available anywhere?

elliottslaughter:
I rebased this as #352 and am planning on merging that PR as soon as it passes. I dropped GNU Make support for now; if someone really wants it we can add it back again.

Thanks again to @ProfFan for all the hard work!

elliottslaughter:
Merged via #352.

PR #315, opened 2018-10-23
Terra building on Windows with VS2015 + Clang6.0.1

Mx7f:
Currently building using llvm+clang compiled from source. Needs testing + appveyor update before merging into master.

Changes:
- LLVM version 6.0.1 (version numbers and path)
- New LLVM lib list
- VS2015 (mostly replacing 120 with 140)
- Needed to add `dbghelp.lib` and `Version.lib` to library list to compile
- `msvcr140.dll` doesn't exist, so I removed it from the the dlls copied into the terra executable directory

With cuda tests skipped I get:

= FAILING tests
dynlib.t
includec.t

520 tests passed. 2 tests failed.


Mx7f:
Not sure why travis-ci is failing; all of my changes are localized to the msvc/ folder.

I have a build of llvm+clang I'd like to put up somewhere so I can test appveyor; it follows the appveyor naming convention: `llvm-6.0.1-windows-amd64-msvc14-msvcrt.7z`. @elliottslaughter, if I send it to you somehow (whats a good way to send a quarter gig file anyways?) would you be willing to put it on your llvm-package-windows page?

Trying to run with CUDA 9.2 gives some errors I'm not quite sure how to debug. Most of the cuda tests fail with something like 

`.\cudahello.t:21: cuModuleLoadData: cuda reported error 218: a PTX JIT compilation failed`

Though some (like `cudaatomic.t` dump what looks like SASS for all of the cuda standard library before doing that:

```
; Function Attrs: alwaysinline inlinehint
define float @__nv_cosf(float %a) local_unnamed_addr #5 {
__nv_isinff.exit.i:
  %result.i.i.i = alloca [7 x i32], align 4
  %0 = tail call float @llvm.fabs.f32(float %a)
  %1 = fcmp oeq float %0, 0x7FF0000000000000
  br i1 %1, label %2, label %__nv_fabsf.exit.i.i

; <label>:2:                                      ; preds = %__nv_isinff.exit.i
  %3 = fmul float %a, 0.000000e+00
  br label %__nv_fabsf.exit.i.i
```

Current status:

```
=================
= FAILING tests
=================
cudaagg.t
cudaaggregate.t
cudaatomic.t
cudaconst2.t
cudaglobal.t
cudahello.t
cudaoffline.t
cudaoo.t
cudaprintf.t
cudashared.t
cudatest.t
cudatex.t
dynlib.t
includec.t
=================

508 tests passed. 14 tests failed.
```

elliottslaughter:
The failing Travis tests seem to be due to llvm.org either being down or else specific items having gone missing. I'm retrying those now and hopefully they'll pass this time.

Mx7f:
Good, Travis is passing. It seems like only `includec.t` is blocking the appveyor build as well. The problem there is `size_t` is not available for some reason. Surprisingly `rsize_t`, a typedef of `size_t`, is imported just fine. I will explore more.

Mx7f:
Narrowing down the problem with includec: `size_t` is processed by the IncludeCVisitor, but is discarded on line: [https://github.com/zdevito/terra/blob/master/src/tcwrapper.cpp#L342](https://github.com/zdevito/terra/blob/master/src/tcwrapper.cpp#L342) because it fails `TD == TD->getCanonicalDecl()`. It is the *only* typedef in the `includec` test to fail this check and if I remove this check then all tests pass.

This feels like a necessary check; any idea what goes horribly wrong if this is removed?

Mx7f:
Ok, will have to put this on hold for a few days after this, but here's what I've tracked down so far:

terra seems to use the wrong headers (it is using VS 12.0 headers, when it should be using VS 14.0). When I augement tcwrapper to dump the files it opens I get:

```
PS D:\Projects\terra\tests> ..\release\bin\terra.exe .\includec.t
openFileForRead(.\mytest.h)
openFileForRead(.\stdio.h)
openFileForRead(C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC/INCLUDE\stdio.h)
openFileForRead(.\crtdefs.h)
openFileForRead(C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC/INCLUDE\crtdefs.h)
openFileForRead(.\sal.h)
openFileForRead(C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC/INCLUDE\sal.h)
openFileForRead(.\ConcurrencySal.h)
openFileForRead(C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC/INCLUDE\ConcurrencySal.h)
openFileForRead(.\vadefs.h)
openFileForRead(C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC/INCLUDE\vadefs.h)
```

but, crucially, this is *not* happening when I run clang:

```
D:\Projects\llvm-6.0.1.src\build\install\bin>clang.exe main.c -v
clang version 6.0.1 (tags/RELEASE_601/final)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: D:\Projects\llvm-6.0.1.src\build\install\bin
 "D:\\Projects\\llvm-6.0.1.src\\build\\install\\bin\\clang.exe" -cc1 -triple x86_64-pc-windows-msvc19.0.24215 -emit-obj -mrelax-all -mincremental-linker-compatible -disable-free -disable-llvm-verifier -discard-value-names -main-file-name main.c -mrelocation-model pic -pic-level 2 -mthread-model posix -fmath-errno -masm-verbose -mconstructor-aliases -munwind-tables -target-cpu x86-64 -dwarf-column-info -debugger-tuning=gdb -momit-leaf-frame-pointer -v -resource-dir "D:\\Projects\\llvm-6.0.1.src\\build\\install\\lib\\clang\\6.0.1" -internal-isystem "D:\\Projects\\llvm-6.0.1.src\\build\\install\\lib\\clang\\6.0.1\\include" -internal-isystem "C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\include" -internal-isystem "C:\\Program Files (x86)\\Windows Kits\\10\\Include\\10.0.14393.0\\ucrt" -internal-isystem "C:\\Program Files (x86)\\Windows Kits\\10\\include\\10.0.14393.0\\shared" -internal-isystem "C:\\Program Files (x86)\\Windows Kits\\10\\include\\10.0.14393.0\\um" -internal-isystem "C:\\Program Files (x86)\\Windows Kits\\10\\include\\10.0.14393.0\\winrt" -fdebug-compilation-dir "D:\\Projects\\llvm-6.0.1.src\\build\\install\\bin" -ferror-limit 19 -fmessage-length 120 -fno-use-cxa-atexit -fms-extensions -fms-compatibility -fms-compatibility-version=19.0.24215 -fdelayed-template-parsing -fobjc-runtime=gcc -fseh-exceptions -fdiagnostics-show-option -fcolor-diagnostics -o "C:\\Users\\mmara\\AppData\\Local\\Temp\\main-d2f7e6.o" -x c main.c
clang -cc1 version 6.0.1 based upon LLVM 6.0.1 default target x86_64-pc-windows-msvc
#include "..." search starts here:
#include <...> search starts here:
 D:\Projects\llvm-6.0.1.src\build\install\lib\clang\6.0.1\include
 C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include
 C:\Program Files (x86)\Windows Kits\10\Include\10.0.14393.0\ucrt
 C:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\shared
 C:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\um
 C:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\winrt
End of search list.
 "C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\bin\\amd64\\link.exe" -out:a.exe -defaultlib:libcmt "-libpath:C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\lib\\amd64" "-libpath:C:\\Program Files (x86)\\Windows Kits\\10\\Lib\\10.0.14393.0\\ucrt\\x64" "-libpath:C:\\Program Files (x86)\\Windows Kits\\10\\Lib\\10.0.14393.0\\um\\x64" -nologo "C:\\Users\\mmara\\AppData\\Local\\Temp\\main-d2f7e6.o"
```

See how it properly uses ` C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include`.

I can't find any hardcoding of these include paths in terra from a quick search, so I'm at a bit of a loss. I'll dig deeper when I get a chance.

elliottslaughter:
Thanks for doing this. It's very encouraging that things are working as well as they are!

I tried to get some LLVM builds going on AppVeyor, because I've had the experience before that subtle differences in the machine where LLVM and Terra are being built can cause issues. However, all my builds have timed out. So I'm not sure what to do about that except beg the AppVeyor folks for even longer timeouts.

Anyway, that's here:

https://github.com/elliottslaughter/llvm-package-windows/tree/clang-6.0.x

https://ci.appveyor.com/project/elliottslaughter/llvm-package-windows/history

I unfortunately don't have a good understanding of how the system include paths are set. I'd been under the impression that they got baked into the Clang library at build time, but it's entirely possible I'm wrong about that. But I've definitely had issues taking Terra builds and moving them to another machine where the libraries are in a different place.

I tried doing a grep through your built LLVM/Clang binary, and didn't have any luck finding those include paths embedded anywhere. I believe grep should be able to find them, even if they are in the middle of a binary, unless the encoding is very strange. I suppose I could try to look for UTF16 encoded strings instead of UTF8; I'm not quite sure how to configure grep to do that.

I guess another possibility is it's just auto-detecting the oldest MSVC version? My understanding is that OS image has both 2015 and 2013 installed.

Mx7f:
Found some fun in terralib.lua:

```
terra.systemincludes = List()
if ffi.os == "Windows" then
    -- this is the reason we can't have nice things
    local function registrystring(key,value,default)
    	local F = io.popen( ([[reg query "%s" /v "%s"]]):format(key,value) )
		local result = F and F:read("*all"):match("REG_SZ%W*([^\n]*)\n")
		return result or default
	end
	terra.vshome = registrystring([[HKLM\Software\WOW6432Node\Microsoft\VisualStudio\12.0]],"ShellFolder",[[C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\]])
	local windowsdk = registrystring([[HKLM\SOFTWARE\Wow6432Node\Microsoft\Microsoft SDKs\Windows\v8.1]],"InstallationFolder",[[C:\Program Files (x86)\Windows Kits\8.1\]])	

	terra.systemincludes:insertall {
		("%sVC/INCLUDE"):format(terra.vshome),
		("%sVC/ATLMFC/INCLUDE"):format(terra.vshome),
		("%sinclude/shared"):format(windowsdk),
		("%sinclude/um"):format(windowsdk),
		("%sinclude/winrt"):format(windowsdk),
		("%s/include"):format(terra.cudahome)
	}
```

This ends up being inserted before the clang include paths (which contain the VS 14.0 directories and the windows 10 sdk).

Here are the paths added from that monstrosity, followed by the ones added from reading clang:

```
terra.systemincludes: C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC/INCLUDE
terra.systemincludes: C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC/ATLMFC/INCLUDE
terra.systemincludes: C:\Program Files (x86)\Windows Kits\8.1\include/shared
terra.systemincludes: C:\Program Files (x86)\Windows Kits\8.1\include/um
terra.systemincludes: C:\Program Files (x86)\Windows Kits\8.1\include/winrt
terra.systemincludes: C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v9.2/include
terra clangpath: C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/INCLUDE
terra clangpath: C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/ATLMFC/INCLUDE
terra clangpath: C:/Program Files (x86)/Windows Kits/10/include/10.0.14393.0/ucrt
terra clangpath: C:/Program Files (x86)/Windows Kits/NETFXSDK/4.6.1/include/um
terra clangpath: C:/Program Files (x86)/Windows Kits/10/include/10.0.14393.0/shared
terra clangpath: C:/Program Files (x86)/Windows Kits/10/include/10.0.14393.0/um
terra clangpath: C:/Program Files (x86)/Windows Kits/10/include/10.0.14393.0/winrt
terra clangpath: /include
```
The cuda include seems important, but I don't know why we are doing the registry parsing for the rest.

Mx7f:
Seems like getting the right headers was a red herring anyways. Even switching the order of the clang and terra include path order, so we read from the right headers, we still get the problem of 
the `size_t` typedef failing the IsCanonical() check. I'll refocus there.

Mx7f:
I think I found the ultimate culprit: `Sema::Initialize` in clang has this code snippet:

```
  // Initialize Microsoft "predefined C++ types".
  if (getLangOpts().MSVCCompat) {
    if (getLangOpts().CPlusPlus &&
        IdResolver.begin(&Context.Idents.get("type_info")) == IdResolver.end())
      PushOnScopeChains(Context.buildImplicitRecord("type_info", TTK_Class),
                        TUScope);

    addImplicitTypedef("size_t", Context.getSizeType());
  }
```

So when compiling to be MSVC compatible, clang has a sourceless typedef as the canonical typedef of `size_t`. This is the *only* non-OpenCL internal typedef within clang; which leads me to think we should special case this.

But the confusing thing to me is that clang 3.5.2 has the same line, so I don't know how terra didn't have this problem before. Compiling clang 3.5.2 from source so I can diff...

Mx7f:
Ok, more progress: with clang 3.5.2, the relevant code is:

```
  if (PP.getLangOpts().MSVCCompat && PP.getLangOpts().CPlusPlus) {
    if (IdResolver.begin(&Context.Idents.get("type_info")) == IdResolver.end())
      PushOnScopeChains(Context.buildImplicitRecord("type_info", TTK_Class),
                        TUScope);
    addImplicitTypedef("size_t", Context.getSizeType());
  }
```

The difference being `PP.getLangOpts().CPlusPlus` was at some point (a quick check reveals this change happened in in 3.6) moved to the inner condition from the outer.  `PP.getLangOpts().CPlusPlus` is false when called by terra, so the Implicit typedef is never added in 3.5.2 (but is in all subsequent versions).

I believe this means the proper fix is to handle `size_t` specially within terra in Windows. I'll come up with a patch soon.

An unrelated problem, but one I want to solve before we merge: 
My local compilation of clang/LLVM 3.5.2 has a superset of the libraries used in terra's binary release:
```
gtest.lib
gtest_main.lib
LLVMAArch64AsmParser.lib
LLVMAArch64AsmPrinter.lib
LLVMAArch64CodeGen.lib
LLVMAArch64Desc.lib
LLVMAArch64Disassembler.lib
LLVMAArch64Info.lib
LLVMAArch64Utils.lib
LLVMARMAsmParser.lib
LLVMARMAsmPrinter.lib
LLVMARMCodeGen.lib
LLVMARMDesc.lib
LLVMARMDisassembler.lib
LLVMARMInfo.lib
LLVMCppBackendCodeGen.lib
LLVMCppBackendInfo.lib
LLVMHexagonAsmPrinter.lib
LLVMHexagonCodeGen.lib
LLVMHexagonDesc.lib
LLVMHexagonInfo.lib
LLVMMipsAsmParser.lib
LLVMMipsAsmPrinter.lib
LLVMMipsCodeGen.lib
LLVMMipsDesc.lib
LLVMMipsDisassembler.lib
LLVMMipsInfo.lib
LLVMMSP430AsmPrinter.lib
LLVMMSP430CodeGen.lib
LLVMMSP430Desc.lib
LLVMMSP430Info.lib
LLVMPowerPCAsmParser.lib
LLVMPowerPCAsmPrinter.lib
LLVMPowerPCCodeGen.lib
LLVMPowerPCDesc.lib
LLVMPowerPCDisassembler.lib
LLVMPowerPCInfo.lib
LLVMR600AsmPrinter.lib
LLVMR600CodeGen.lib
LLVMR600Desc.lib
LLVMR600Info.lib
LLVMSparcAsmParser.lib
LLVMSparcAsmPrinter.lib
LLVMSparcCodeGen.lib
LLVMSparcDesc.lib
LLVMSparcDisassembler.lib
LLVMSparcInfo.lib
LLVMSystemZAsmParser.lib
LLVMSystemZAsmPrinter.lib
LLVMSystemZCodeGen.lib
LLVMSystemZDesc.lib
LLVMSystemZDisassembler.lib
LLVMSystemZInfo.lib
LLVMXCoreAsmPrinter.lib
LLVMXCoreCodeGen.lib
LLVMXCoreDesc.lib
LLVMXCoreDisassembler.lib
LLVMXCoreInfo.lib
```

The gtest libraries are note used or needed for linking, but linking fails if I don't specify the other libraries. @elliottslaughter Is this a problem? Should I be somehow disabling these libraries from compilation (seems like they are for cross-compilation, but I don't know what CMake flag controls whether we use them).

Mx7f:
Compiling LLVM 7.0 while waiting: I'm guessing the extra libraries above are coming from CMake choosing

`LLVM_TARGETS_TO_BUILD` all,

which causes:
```
Targeting AArch64
Targeting AMDGPU
Targeting ARM
Targeting BPF
Targeting Hexagon
Targeting Lanai
Targeting Mips
Targeting MSP430
Targeting NVPTX
Targeting PowerPC
Targeting Sparc
Targeting SystemZ
Targeting X86
Targeting XCore
```
I'm guessing the builds on https://github.com/elliottslaughter/llvm-package-windows/releases/ set `LLVM_TARGETS_TO_BUILD` to `X86;NVPTX`, is that right?

elliottslaughter:
Right, if you want to build a subset of LLVM libraries you can do it like:

https://github.com/elliottslaughter/llvm-package-windows/blob/clang-6.0.x/set-env.bat#L130

Mx7f:
The windows build works and is passing appveyor (though I haven't tested CUDA yet). I have binary builds of llvm+clang with ptx support up on my fork of llvm-package-windows (https://github.com/Mx7f/llvm-package-windows/releases/) for versions 4,5,6, and 7, with the proper library lists in the Makefile (4 and 7 are not part of the terra appveyor matrix since they are currently unsupported).

I do not know why travis is failing (only on expvec.t in /usr/bin/../lib/gcc/x86_64-linux-gnu/4.8/include/immintrin.h, and only for llvm 3.5).

elliottslaughter:
Am I correct in saying that the Terra system includes will still be wrong in this case? I realize we can "fix" this by putting Clang first, but it seems we ought to fix whatever auto-detect logic we've got. If that means using a define to tell us what MSVC version we're being built with, so that we can look up the right registry key or whatever, that seems preferrable to having this inconsistency between the include paths in Windows and everything else.

Mx7f:
Yes, that would be preferable. `_MSC_VER` is used elsewhere in terra in cpp code; what would be a good way to pipe it to the lua code in `terralib.lua`?

elliottslaughter:
I'm not sure if it's the "best" way, but you can see here for an example of how to pass through values from C++ to Terra: https://github.com/zdevito/terra/pull/314/files#diff-cad54c5093af7b65edb23532e668e00c

elliottslaughter:
I spent some time trying to get this cleaned up today, and so far everything I've tried hasn't worked. It seems the values change in non-trivial ways with the windows versions, or maybe we're not doing things the "right" way for the platform.

I also tried just getting rid of Terra's Windows path settings (why should we duplicate them anyway?), but somehow that didn't work either.

Anyway, at this point I'm inclined to just take this as-is for the moment, unless @Mx7f has an interest in spending more time on this.

elliottslaughter:
I've gone ahead and merged this, but am still happy to take commits to do further cleanup if @Mx7f or anyone else is interested in putting in the time.

@Mx7f Thanks so much for doing this!

Mx7f:
I will be trying to make this work over the next few days... I need to be able to run Opt on Cuda 9 for some comparisons...

PR #317, opened 2018-11-09
Define terralib.os for forwards-compatibility with PUC Lua

elliottslaughter:
This makes master compatible with the develop branch, so users can start using `terralib.os` now to avoid unnecessary dependencies on the `ffi` library.


elliottslaughter:
I've been thinking about this more, and I think we can support PUC Lua without needing to get rid of `ffi` entirely.

PR #322, opened 2018-11-22
CMake support for Windows

elliottslaughter:
This pull request adds (or will add, it isn't quite there yet) Windows support for the CMake build.

At the moment it fails because some of the dependencies aren't being set up correctly.


elliottslaughter:
I'm hoping to solicit some help from people familiar with CMake. (CC: @OvermindDL1 @Clyybber) For some reason the dependencies in the LuaJIT build aren't being properly respected, and I can't for the life of me figure out what's wrong.

The structure is basically:

```
add_custom_command(
    OUTPUT ${LUAJIT_EXECUTABLE}
    COMMAND ...
)
add_custom_command(
      OUTPUT "${LUAJIT_INSTALL_PREFIX}/lua/jit/${LUAJIT_SOURCE_NAME}"
      DEPENDS ${LUAJIT_EXECUTABLE}
      COMMAND ...
)
add_custom_target(
  LuaJIT
  DEPENDS
    ${LUAJIT_EXECUTABLE}
    "${LUAJIT_INSTALL_PREFIX}/lua/jit/${LUAJIT_SOURCE_NAME}"
)
add_custom_command(
  OUTPUT ${PROJECT_BINARY_DIR}/clangpaths.h
  DEPENDS LuaJIT
  COMMAND ...
)
add_custom_target(
  TerraGeneratedFiles
  DEPENDS
    ${PROJECT_BINARY_DIR}/clangpaths.h
)
```

On the first build, `LUAJIT_EXECUTABLE` gets built, but not `"${LUAJIT_INSTALL_PREFIX}/lua/jit/${LUAJIT_SOURCE_NAME}"`. So the later command, which depends on the `LuaJIT` target, fails. But *on my second build*, the `"${LUAJIT_INSTALL_PREFIX}/lua/jit/${LUAJIT_SOURCE_NAME}"` file gets created properly, and everything builds.

I understand there is some subtlety about what kind of things are allowed to depend on what, but I think I'm following all the rules here. At least I can't see anything wrong.

If you want to see the actual sources, you can find them here:

https://github.com/elliottslaughter/terra/blob/0e61b66e2d044dadbeba9c833b2bc13d400eda76/cmake/Modules/GetLuaJIT.cmake

https://github.com/elliottslaughter/terra/blob/0e61b66e2d044dadbeba9c833b2bc13d400eda76/src/CMakeLists.txt

Thanks in advance for any guidance.

Clyybber:
This is really wierd. It looks correct to me. Perhaps we should ask some CMake guru's on stackoverflow or IRC?

elliottslaughter:
StackOverflow post here: https://stackoverflow.com/q/53435714/188046

I've worked around it for the moment by issuing the problematic commands at configure time, which is ugly, but seems to get the job done. I would still like to know why the original version didn't work though, since I don't know when I might hit this in the future.

OvermindDL1:
As I recall (it's been a long long time since I dealt with VS) Visual Studio has an oddity about how dependencies can be mapped, essentially it has two types of dependencies, and the 'main' dependency can only be a singular dependency, so I'd change this:
```cmake
add_custom_target(
  LuaJIT
  DEPENDS
    ${LUAJIT_EXECUTABLE}
    "${LUAJIT_INSTALL_PREFIX}/lua/jit/${LUAJIT_SOURCE_NAME}"
)
```
To this:
```cmake
add_custom_target(
  LuaJIT
  DEPENDS "${LUAJIT_INSTALL_PREFIX}/lua/jit/${LUAJIT_SOURCE_NAME}"
)
```
As `${LUAJIT_INSTALL_PREFIX}/lua/jit/${LUAJIT_SOURCE_NAME}` depends on `${LUAJIT_EXECUTABLE}` anyway (Visual Studio's "main" command dependency can only be a single dependency, not multiple, it's a VS annoyance...).

However, this whole luajit setup looks odd...  Why not use the luadist luajit CMake setup at https://github.com/LuaDist/luajit instead?

elliottslaughter:
Ok, got this to the point where it builds, links, and runs most of the test suite. (Currently 8 failing tests.)

blackhole12:
This is still failing all the Visual Studio tests, and it is failing each one for unique reasons. Visual Studio 2013 seems impressively broken in ways I cannot begin to fathom. Visual Studio 2015 can't decide if there's a linker error or if half the tests should just fail for absolutely no reason. I will have to attempt building this with Visual Studio 2019 to see which errors are not simply environment issues.

elliottslaughter:
For CMake, I'd be willing to drop Visual Studio 2013 support altogether and focus only on the more recent versions. I think we're stuck with 2015 for the moment since that was what LLVM was built with, but even there we could rebuild with 2017 or 2019 if it would make things easier.

blackhole12:
Re: the LLVM build issue, I feel obligated to point out that vcpkg #387 can, in fact, build LLVM from source on windows, but it's LLVM 8. A proper vcpkg implementation would create dependencies on both LLVM and LuaJIT and build them from source using the correct triplets.

It would also be possible to statically compile the runtime instead of manually copying msvc140.dll, which you aren't supposed to do in the first place. If you want Terra to be portable, the correct way is to use `/MT` on everything, which removes the dependency entirely. [I already figured out how to convince LLVM to compile this way](https://github.com/innative-sdk/innative/blob/master/build-llvm.ps1#L57), so I could use the same method for Terra.

blackhole12:
After I managed to recreate these tests on my machine, `multiterra.t` is actually working but has spurious output from LLD, so I've added it to the skip list.

dynlib.t explodes with over 350 unresolved references in a spectacular failure probably resulting from not linking something properly

All 5 other tests are a result of not being able to find `std.t` or `parsing.t`, which normally lie in the `lib` folder. For whatever reason, this build of terra doesn't seem to be searching through those folders and I have no idea why. Moving `std.t` and `parsing.t` to the `include` directory lets all these tests pass.

PR #327, opened 2018-12-11
Add tests for external projects

elliottslaughter:
Current status:

 - [ ] Regent
 - [ ] Rigel
 - [ ] Opt

Others?


aiverson:
I am currently working on updating quicksand for the current version of Terra. When I have it running, should that become part of this test package, and what should I do with regards to that?

elliottslaughter:
Yes, if you have significant pieces of Terra code (ideally with tests), let's include them as external tests on the CI. The goal of this effort is to have as broad test coverage as possible, so that we can have some degree of assurance that we aren't breaking the ecosystem when we make changes.

Depending on whether you get to it before or after this PR merges, we could either add it here or as a new PR.

aiverson:
I don't anticipate getting it here before this PR merges. Quicksand has a considerable amount of nontrivial code in it, and large parts need major rewrites to fit the changes to the language since it was last maintained.

elliottslaughter:
@Mx7f @ProfFan Is there a way to build Opt without CUDA support? Otherwise I'm not sure I have any realistic hope of testing it in CI. (I can install CUDA, but there will never be a GPU available.)

ProfFan:
@elliottslaughter I think Opt is GPU only.

Mx7f:
@elliottslaughter  Yes, unfortunately Opt is currently without non-CUDA support.  CI would be invaluable, and this limitation has come up before; so I'm seriously considering writing a small CUDA-emulation shim for the CPU (not built initially for performance) to have in Opt.

PR #330, opened 2019-01-02
build on aarch64 (nvidia xavier)

cinterloper:
This is the minimum change required to build on NVIDIA Xavier ( linux4tegra on aarch64 ) in addition to setting the luajit version to 2.1 ( I used beta3 )
When using CMAKE I also have to disable LLVM_ENABLE_RTTI 

I have not yet been able to get CUDA (10) to work, even though it appears to build with cuda when the makefile.inc is configured.

### DIFF
--- a/src/tdebug.cpp
+++ b/src/tdebug.cpp
@@ -3,7 +3,7 @@
 #include "terrastate.h"
 #include "tcompilerstate.h"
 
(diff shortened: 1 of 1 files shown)


elliottslaughter:
Thanks for doing this; looks good!

PR #337, opened 2019-01-22
Workaround for Skylake-Clang 3.8 Bug

gilbo:
On Skylake processors, Clang 3.8.x generates illegal AVX instructions.  Therefore we disable AVX code generation entirely in that situation.  More information can be found on this bug discussion:
https://www.bountysource.com/issues/37851678-llvm-3-8-1-compiles-vmovsd-on-intel-skylake-using-evex


elliottslaughter:
Thanks!

Issue #324, opened 2018-11-28
Syntax for adding annotations and metadata to fields of structs


aiverson:
It is possible to add arbitrary key/value pairs as extra information onto structs and fields of structs using the Lua API. Adding data to a struct with the Lua API is very straightforward, as is adding it to an entry in a struct which was created with the Lua API. However, adding data to an entry of a struct created using the syntactic sugar is inconvenient. The obvious tool to do this is to have a bit of syntax for adding arbitrary key/values to structs in the syntax for struct declarations.

Having such metadata is useful for a variety of applications, for example in a serialization library to indicate whether an int8 should be serialized as a number or a single character string.

 I have been unable to locate such a syntax in the documentation or tests and would like to write the extension if it doesn't exist yet.

Do any more active developers have suggestions or preferences for this syntax before I start implementing it, or would it not fit well as a part of the core of Terra?

aiverson:
Possible syntax:
```struct example {
    fieldA: int64 @JSONserialized("string")
    fieldB: int8 @{jsonbehavior = "char", csvbehavior = "char"}
    fieldC: int @[boundschecked(-10, 10), lazy(`fieldA / fieldB / fieldB)]
}
```
A type declaration followed by an @ by analogy to python's decorators and java's annotations starts a metatagging expression. A meta expressions may start with a name, in which case the expression evaluates to tagging data, a open curly brace, in which case the table literal is used as tagging data that is merged into the field data as part of the metatype. A square bracket starts a comma separated list of tagging expressions which are merged or composed into the final tagging.

This seems like a workable syntax.

elliottslaughter:
Just as a counterpoint, what about a syntax like:

```
struct example [whatever] {
  fieldA [something] : int64,
  fieldB [another] : int8,
  fieldC [thing] : int,
}
```

The main advantage is that `[]` is already the escape operator, so it's hopefully relatively obvious that you can use arbitrary Lua syntax here. Because it's on the side of the name (vs the type) it should parse unambiguously. It doesn't involve introducing any novel syntax, only a new position where escapes can be introduced. (You want a list? Just write `fieldName [{a, b, c}]` like you do in a normal escape.)

Personally, I don't have a super strong need for this. I'm sure @aiverson is aware, but just for anyone else reading this, you can already do:

```
struct example { ... }
example.annotation = something
example:getentries()[1].annotation = something
```

Or:

```
local example = terralib.types.newstruct("example")
example.entries = {
[...]

PR #346, opened 2019-02-10
Added annotations to individual fields in a struct #324

aiverson:
Initial implementation of #324 
Multiple comma separated annotations are not yet supported.


aiverson:
Does anyone know what is broken on AppVeyor here?

elliottslaughter:
I'm not sure why AppVeyor is failing, but I'd recommend looking at the changes to `genclangpaths.lua`/`geninternalizedfiles.lua` since those are involved in generating header files that will be imported into C code. (By the way, why are these changes required?) I suppose it could also be in `lparser.cpp`, but I'm having a hard time seeing how that change is related. In theory, it should be impossible for changes to the other Lua/Terra files to result in compile errors like this.

aiverson:
The change in geninternalizedfile.lua makes it not fail to compile when it doesn't have any files to internalize.

The change in genclangpaths.lua makes it so that it can compile when the clang path has a space in it, for instance "C:\Program Files\Clang\clang.exe"

I encountered both of these problems while trying to build from source on windows, but eventually gave up on trying to get my windows system configured to build it.

elliottslaughter:
Maybe we should split those into a different PR? They seem like reasonable changes, but for this specific PR it'll be easier to debug if there's less to look at.

aiverson:
Sure, I'll revert that commit and try again.

aiverson:
That didn't magically fix it.

elliottslaughter:
Oh, please set the PR to be based on `master`. `develop` is in a strange state and I have no idea if it builds at all in AppVeyor.

aiverson:
Oh. That could be the problem. I'll put it on master.

aiverson:
Is there any chance of putting it in a good state soon?

elliottslaughter:
I don't think we're ever going to merge develop in its entirety. The parts that have been useful we've been pulling piecemeal into master. Other parts we've taken a fairly different approach on (PUC Lua support) which results in substantially less churn vs what had been done on develop. There are still some parts that would be nice to have (LLVM 3.8 debug symbols), but I haven't put in the effort to extract them.

elliottslaughter:
Closing since this is superseded by #347 (if I understand correctly---please let me know if I'm misunderstanding).

Issue #288, opened 2018-06-25
cuda error 500 on CUDA 9.0 and LLVM 3.8


elliottslaughter:
Reposting a bug report from @chichengumich in [this comment](https://github.com/zdevito/terra/pull/278#issuecomment-399790307):

> Hi, I was trying to compile terra with CUDA 9.0 and llvm 3.8. The compliation went through without any error. However, when I tried to run cudahello.t, it returns
> ```
> cudahello.t:21: cuModuleGetFunction: cuda reported error 500
> stack traceback:
> 	[C]: in function 'error'
> 	[string "<string>"]:247: in function 'cudacompile'
> 	cudahello.t:21: in main chunk
> ```
> Is there any solution?

elliottslaughter:
@chichengumich Can you please reply with enough information so that we can reproduce your bug? What OS and OS version? C++ compiler and version? How did you obtain LLVM (build from source, binary (if so from where?), etc.).

Also, can you confirm if your machine has a GPU attached?

I asked @manopapad to reproduce your bug, and we weren't able to do it with either CUDA 8.0 or 9.0. So it would appear that there's something specific to your setup that is problematic.

Thanks.

chichengumich:
@elliottslaughter Thank you for your response!
I am using ubuntu 16.04 LTS.
For compiler I have clang 3.8, 3.9 and 6.0 installed. The default is clang 3.8:
They are all installed using `sudo apt-get install clang-3.8` and then `sudo apt-get install libclang-3.8-dev`.
For CUDA, I have CUDA 8.0, 9.0 and 9.1 installed. CUDA_HOME is set to 
```
$ echo $CUDA_HOME
/usr/local/cuda-9.0
$ ls -la /usr/local | grep cuda
lrwxrwxrwx  1 root root   19 Jun 24 16:32 cuda -> /usr/local/cuda-8.0
drwxr-xr-x 17 root root 4096 Jun 24 16:32 cuda-8.0
drwxr-xr-x 18 root root 4096 Feb 26 10:15 cuda-9.0
drwxr-xr-x 19 root root 4096 Feb 26 10:05 cuda-9.1

```
I have a GTX1070 attached, with my regular CUDA program compiles and running correctly. I also have a copy of Opt installed in a separate `Optlang/Opt` and `Optlang/terra' installed. In which terra is installed [terra-2016-03-25](https://github.com/zdevito/terra/releases/download/release-2016-03-25/terra-Linux-x86_64-332a506.zip)
```
[...]

Issue #199, opened 2016-08-09
Issue to be aware when calling terralib.saveobj with a target


xspager:
This example reproduce the issue I was having while I was trying to generate code without SSE instructions. This code:

```terra
local target = terralib.newtarget {
  --Triple = "x86_64-apple-darwin";
  Triple = "x86_64-pc-linux";
  -- CPU = "";
  Features = "";
}

local C = terralib.includecstring([[
#include <stdio.h>
typedef struct {
    int a;
    int b;
} foo;
]], {})

terra main()
    var v: C.foo
	v.a = 1
	v.b = 2

    C.printf("%d %d\n", v.a, v.b)
end

terralib.saveobj("target", {main=main}, {}, target)
```
Results in the error: 
``terra: src/tcompiler.cpp:553: llvm::StructType* Types::CreateStruct(Obj*): Assertion `df' failed.
Aborted (core dumped)``

It builds ok if I pass the **target** to _includecstring_. If I don't pass the target to both _saveobj_ and _includecstring_ this example will build fine too but this function https://github.com/xspager/hello_efi_from_terra/blob/master/foo.t#L23 would build if I passed the target to _saveobj_ but not to _includecstring_ but after linking it it would't run past line https://github.com/xspager/hello_efi_from_terra/blob/master/foo.t#L31

Sorry If I'm not clear, please feel free to ask me questions. I tried to debug both the machine code terra generated and the code around the failing assertion in **tcompiler.cpp** but it is beyond me.

zdevito:
You definately need to pass the target to both includecstring and saveobj.
C headers have a lot of target-specific definitions, and so we need to know
what target it is when including things. Terra should do a better job
reporting errors when the target of the saveobj and includec are
mismatched, but I haven't gotten around to it yet.

It seems like you get another error if you pass the target to both? Can you
describe what happens then?

xspager:
@zdevito, that would be nice to have. :)

Yes, I think I figure it out now, at first I suspected the problem was the use of SSE instructions but turns out it was the AVX set.

xspager:
@zdevito, just to be clear, my initial problem was solved instructing llvm to not use AVX instructions:
```terra
local target = terralib.newtarget {
	Triple = "x86_64-pc-none";
	--CPU = ;
	Features = "-avx"; -- don't use AVX instructions
}
```
Second, I noticed that if you pass a target to objsave but not to includecstring I get the assertion error, I can reproduce the issue with the example.

[...]

PR #349, opened 2019-02-14
Search for imported structs across all compilation targets

manopapad:
This patch will allow us to import a type once using `terralib.includec` and use it in code generated for all compilation targets (e.g. both CPU and GPU functions). I expect this will solve #199, and help with some of the issues in #288. It will also allow Regent's CUDA codegen to handle imported structs.

### DIFF
--- a/src/tcompiler.cpp
+++ b/src/tcompiler.cpp
@@ -308,8 +308,10 @@ bool HostHasAVX() {
 int terra_inittarget(lua_State * L) {
     terra_State * T = terra_getstate(L, 1);
     TerraTarget * TT = new TerraTarget();
-    TT->nreferences = 1;
-    
+    TT->id = T->targets.size();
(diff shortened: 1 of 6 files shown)


elliottslaughter:
Passes on my Ubuntu 16.04 machine with CUDA 9.2 and both LLVM 3.8 and 6.0. Might be good to get some test coverage of other systems as well.

I'll try to look at the code when I can.

elliottslaughter:
@ProfFan @Mx7f Do one of you have a Windows machine with CUDA? If so could I get you to sanity check this patch to make sure the test suite passes? Thanks.

ProfFan:
Cannot help here, my Windows machine is gaming only :)

Mx7f:
I'll be double checking this stuff tomorrow; I'll open a new pr if it doesn't seem to be working.

Issue #368, opened 2019-05-07
iterator bug


capr:
Looks like there's a bug with the for iterator: it works with `for x in iter_struct do` but behaves strangely with `for x in func_returning_iter_struct() do` in that it sometimes calls `func_returning_iter_struct()` twice. Test case:

```
struct iter {n: int}

iter.metamethods.__for = function(self, body)
	return quote
		[ body(`self.n) ]
		[ body(`self.n) ]
	end
end

terra this_is_called_twice(n: int)
	C.printf'this should only be called once\n'
	return iter{n}
end

terra test()
	for x in this_is_called_twice(5) do
		C.printf('%d\n', x)
	end
end
test()
```

aiverson:
I'll take a look. I bet I can fix this quickly.

aiverson:
It isn't really a bug. The issue is that you are putting self in more than one place.

```
struct iter {n: int}

iter.metamethods.__for = function(self, body)
	return quote
                var self_ = [self]
		[ body(`self_.n) ]
		[ body(`self_.n) ]
	end
end

terra this_is_called_twice(n: int)
	C.printf'this should only be called once\n'
	return iter{n}
end

terra test()
	for x in this_is_called_twice(5) do
		C.printf('%d\n', x)
	end
end
test()
```
Should not exhibit the problem.

It is using a splice in more than one spot of the same expression, which is the expression from the for loop. I can modify the code responsible to synthesize a caching variable inside the forlist handler in the compiler, and have a way to do that if you would like.

aiverson:
It's unexpected and undocumented behavior, so I consider it a bug, but I think that it could be fixed either by documenting it or by changing the behavior. I've been writing code to change the behavior, but I'm willing to listen if people think it should be left as is and documented.

capr:
I vote strongly for changing the behavior.

elliottslaughter:
Not that I disagree, but just to play devil's advocate, why is the double evaluation not intended behavior? With any other sort of macro, you'd expect to get the expression as an argument directly and it would be up to the user to manage whether it should be evaluated once or multiple times.

Are there any scenarios that can't be implemented if we get rid of the ability to do multiple evaluation here?

OvermindDL1:
[...]

PR #377, opened 2019-06-22
isolating changes for the for-in variable synthesis

aiverson:
I think I managed to grab all the changes for the variable synthesis and nothing else.
This fixes a bug where accessing the self parameter of an iterator multiple times can cause unexpected behavior.


aiverson:
This addresses #368

elliottslaughter:
Here's another test I'd want to work before we merge this:

[forlist4.t.txt](https://github.com/terralang/terra/files/4094587/forlist4.t.txt)

Issue #372, opened 2019-06-04
Bug in copy assignments


magnatelee:
Current status:

 - [X] (pass in master) `tests/bug372.t`
 - [X] (pass in master) `tests/bug372b.t`
 - [X] (pass  in #391) `tests/bug372c.t` ([gist](https://gist.github.com/elliottslaughter/321fd0969aaabb5e47f8434e0a533525))
 - [X] (pass in #391) `tests/bug372d.t` ([gist](https://gist.github.com/elliottslaughter/e7adba28de2dbc43378de1fdba9766d6))
 - [X] (fast in #391) `tests/bug372_perf.t` ([gist](https://gist.github.com/elliottslaughter/b1f9e22af7d9deedcc02a2b6161c700d))

Original reproducer:

schema.h
```
#ifndef __SCHEMA_H__
#define __SCHEMA_H__

struct Config {
  union  {
    char FileDir[16];
    struct  {
      unsigned int length;
      double values;
    } s;
  } u;
};

#endif // __SCHEMA_H__
```

reproducer.t
```
local C = terralib.includec("stdio.h")
local SCHEMA = terralib.includec("schema.h")

terra main()
  var config : SCHEMA.Config
  C.snprintf(&(config.u.FileDir[0]), 16, "%s", "String111222333")
  C.printf("orig: %s\n", config.u.FileDir)
  var copy = config
  C.printf("copy: %s\n", copy.u.FileDir)
end

main()
```

This code prints out the following (the exact output might be different on different platforms):
```
orig: String111222333
copy: Stri
```
However, the output should have been this:
```
orig: String111222333
copy: String111222333
```
If I change the type of `length` to any 8-byte type, the code produces a correct output.

I suspect Terra doesn't emit correct copy assignment code when the fields of a union are not aligned.

magnatelee:
In fact, this bug is reproduced purely in Terra. Here is a even simpler reproducer:

```
local C = terralib.includec("stdio.h")

struct st
{
  length : uint;
  values : double;
}

struct Config
{
  union {
    FileDir : int8[16],
    s : st,
  }
}

terra main()
  var config : Config
  C.snprintf(&(config.FileDir[0]), 16, "%s", "String111222333")
  C.printf("orig: %s\n", config.FileDir)
  var copy = config
  C.printf("copy: %s\n", copy.FileDir)
end

main()
```

aiverson:
I can reproduce this on my system.

manopapad:
[...]

PR #389, opened 2019-08-19
Another fix for struct assignment, thanks to Seema Mirchandaney

elliottslaughter:
Again following on the series of fixes for #372.


elliottslaughter:
Closing this in favor of #390, which handles strictly more cases and also fixes a compilation time regression.

PR #392, opened 2019-08-30
Add Code of Conduct

elliottslaughter:
Hi everyone,

I am adding a Code of Conduct to the project. You can see the [rendered text here](https://github.com/elliottslaughter/terra/blob/code_of_conduct/CODE_OF_CONDUCT.md).

Terra is getting more contributions from many different people---and that's a great thing! It also means that there are opportunities for miscommunication. Miscommunication means not just making people feel unwelcome---it means the project as a whole can become inefficient, lose contributors and possibly even grind to a halt. Focusing on effective communication is therefore important to the overall health of the project.

The code of conduct is intended to help in two ways. First, to communicate the ideals we strive for. And second, to identify explicitly what should be done in cases that fall so far out of expected norms that it actually results in project dysfunction. Obviously, the second case is expected to be rare, and the hope is that by enumerating the ideals in the first case, we'll be able to steer clear of any potential issues.

It is an explicit non-goal of the code to be pedantic, legalistic, etc. in these matters. As it says in the code itself, please "take it in the spirit in which it's intended---a guide to make it easier to enrich all of us and the technical communities in which we participate".

Please feel free to reply with questions or comments here, or to terra-mods@googlegroups.com if you have any concerns with making public feedback.

Here's to a successful future for Terra!

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,160 @@
+# Terra Code of Conduct
+
+Like the technical community as a whole, the Terra community is made up of a
+mixture of professionals and volunteers from all over the world, working on
+every aspect of the mission---including mentorship, teaching, and connecting
+people.
+
+Diversity is one of our huge strengths, but it can also lead to
+communication issues and unhappiness. To that end, we have a few ground
+rules that we ask people to adhere to. This code applies equally to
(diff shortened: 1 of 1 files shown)


aiverson:
I'm in favor of having this.

Issue #396, opened 2019-09-13
printstats() on uncompiled function raises useless error message


DarkWiiPlayer:
terra dostuff()
		return 10
	end

	-- dostuff:compile()
	print(dostuff:printstats())

This code produces a useless error message

    src/terralib.lua:561: bad argument #1 to 'pairs' (table expected, got nil)

Which wouldn't help at all in a more complex codebase where it's harder to figure out at which point a function is actually compiled.

It would make sense to either compile the function automatically when `printstats` is called on a yet uncompiled function, or to just return `nil` + error message so as to not crash the entire program.

elliottslaughter:
Seems reasonable, and it shouldn't be that hard to implement if you'd like to take a pass at putting together a PR.

DarkWiiPlayer:
Both options seem trivial to implement; which one would be preferable though?

elliottslaughter:
I'd just make `printstats` call `compile` since it's normally something people would do for debugging so taking the hit (in time to wait for compilation to finish) shouldn't affect any large codes.

DarkWiiPlayer:
Sounds reasonable; my was that compiling a function is a side effect and I personally prefer debug functionality to be free of side effect, so the programs behavior doesn't change from adding / removing a debug output. Correct me if I'm wrong, but after defining a function (and before compiling it) you can change other things that will affect the function, right?

elliottslaughter:
This is less true than it used to be now that we use eager type checking (though possibly we may go back to being a little lazier at some point).

I think the key question is whether `printstats` is something that would be useful even if the function were not compiled? Like the `disas` function is clearly not, so it just compiles the function for you if it isn't already. I'd be inclined to put `printstats` in this category as well.

elliottslaughter:
Merged the PR. Thanks again!

PR #397, opened 2019-09-13
Compile functions when calling `printstats`

DarkWiiPlayer:
fix for #396

### DIFF
--- a/src/terralib.lua
+++ b/src/terralib.lua
@@ -557,6 +557,7 @@ function T.terrafunction:disas()
     terra.disassemble(terra.jitcompilationunit:addvalue(self),self:compile())
 end
 function T.terrafunction:printstats()
+    self:compile()
     print("definition ", self:gettype())
     for k,v in pairs(self.stats) do
         print("",k,v)


elliottslaughter:
Looks good, will merge after CI passes.

elliottslaughter:
I am confused, but this actually seems to be causing a test failure on macOS with LLVM 3.8. I've rerun this job 4 times, and it seems to be deterministic. I also ran a dummy PR on master (#398) to confirm it doesn't show up there. I'm not sure what's going on though because as far as I can tell the failing test (`leaktest.t`) does not even call `printstats`.

DarkWiiPlayer:
Well, I wish I could help with that, but if you say the test doesn't even directly call `printstats`, I highly doubt I'd be of any help figuring out what's causing the problem :\

elliottslaughter:
I'm just going to merge this since `grep` isn't reporting any actual callsites for `printstats`. Perhaps there's an issue with the size of `terralib.lua` or something (since it gets included in the binary as a string), but aside from those sorts of second-order effects I can't think of what would be going on.

If the test keeps failing, I'll just turn it off; LLVM 3.8 isn't a high priority on macOS anyway.

PR #402, opened 2019-09-27
Add support for noreturn functions

sometimes-i-send-pull-requests:
Terra supports marking functions as always-inline, never-inline, and don't-optimize. LLVM also supports marking functions as noreturn. This PR exposes LLVM's noreturn attribute to Terra.


elliottslaughter:
Looks good, thanks!

PR #412, opened 2019-10-12
Get system include paths from clang instead of hardcoding them

### LINKED ISSUE #393: Build failure on Ubuntu 19.04
Build fails with:

```
# extract needed LLVM objects based on a dummy linker invocation
< build/llvm_objects/llvm_list build/bin/luajit-2.0.5 src/unpacklibraries.lua build/llvm_objects
build/bin/luajit-2.0.5: src/unpacklibraries.lua:20: attempt to index local 'archivepath' (a nil value)
stack traceback:
	src/unpacklibraries.lua:20: in main chunk
	[C]: at 0x55e5b2b97040
make: *** [Makefile:223: build/llvm_objects/llvm_list] Error 1
```

Repro steps:

Get an Ubuntu 19.04 shell via Docker:

```
docker run -ti ubuntu:19.04
```

Then run:

```
apt install build-essential git cmake wget llvm-6.0-dev libclang-6.0-dev clang-6.0
git clone https://github.com/zdevito/terra.git
cd terra
LLVM_CONFIG=llvm-config-6.0 CLANG=clang-6.0 make
```

Build works ok with CMake, though you do additionally need to install `libedit-dev zlib1g-dev` (which again for some reason do not seem to get pulled along with the LLVM dependency).

Perhaps another reason to do #383.

blackhole12:
This uses a hack from [Dragon FFI](https://github.com/aguinet/dragonffi/blob/2a205dbe4dd980d5dd53026c871514795573a7fb/lib/dffi_impl.cpp#L139) to extract clang's own system header files, avoiding the necessity of trying to find them ourselves. It then deletes `genclangpaths.lua` and removes `clangpaths.h` from both cmake and makefile builds, fixing the hardcoded paths problem.

This pull request also makes some minor changes to windows-specific hacks that were put in place, removing them for newer versions that no longer require these workarounds to promote code homogeneity. In particular, we no longer use the Visual Studio version to set `fms-compatibility-version`, because this is actually the version of `cl.exe`, not the SDK, or Visual Studio, or anything else. It may be preferable to try and let clang figure this value out on it's own, because [the logic does exist](https://reviews.llvm.org/D20136), but is inaccessible. For now, a value of `1800` is fine for the purposes of parsing most windows headers, and is better than simply using "whatever version of visual studio we happened to compile with".

Incidentally, this fixed my own problems with compiling Terra on Ubuntu 19, so this pull request may have accidentally fixed #393, but someone will need to verify this. I've tested this pull request on both Windows 10 and Ubuntu 19, but we will need to see if I broke any of the test environments, since this is a nontrivial change to the build process.


elliottslaughter:
This is awesome! Thanks for doing this.

There is one glitch I noticed where the CUDA tests seem to be broken by this. I can build, but then CUDA somehow ends up not being put on the system include path. Setting `INCLUDE_PATH` manually avoids the problem, but really seems like an unfortunate compromise given what we're trying to do here.

```
$ ../terra cudaoo.t 
<buffer>:1:10: fatal error: 'cuda_runtime.h' file not found
#include "cuda_runtime.h"
         ^~~~~~~~~~~~~~~~
compilation of included c code failed

stack traceback:
	[C]: in function 'registercfile'
	src/terralib.lua:3412: in function 'includec'
	cudaoo.t:9: in main chunk

$ INCLUDE_PATH=/usr/local/cuda-9.2/targets/x86_64-linux/include ../terra cudaoo.t
```

I assume we're just missing some sort of Clang configuration to pick up all the system include directories. Do you have any interest/availability to look into this?

blackhole12:
What you are expecting is not what Terra is doing, and I don't think Terra actually ever did what you described except by accident. The CUDA header logic was preserved in `terralib.lua`, but it tries to use `$CUDA_HOME$` to find CUDA on Linux.

```
if ffi.os == "Windows" then
  terra.cudahome = os.getenv("CUDA_PATH")
else
  terra.cudahome = os.getenv("CUDA_HOME") or "/usr/local/cuda"
end
```
If you want to _change_ how Terra finds CUDA and try to convince clang itself to look up the include directory, this is hard to test because none of the current tests you have set up actually do things this way. I've pushed a change where I call `AddCudaIncludeArgs`, but I have no idea if this will work and have no test to verify it against.

PR #414, opened 2019-10-21
Fix stdio.h include problem

### LINKED ISSUE #401: Terra Cannot Import Inlined C Functions
As I discovered while trying to get the tests to work again in #400, Terra shouldn't actually be working on windows _at all_. This is because `printf` is defined as an inline function and [has been since Visual Studio 2015](https://social.msdn.microsoft.com/Forums/vstudio/en-US/5150eeec-4427-440f-ab19-aecb26113d31/updated-to-vs-2015-and-now-get-unresolved-external-errors?forum=vcgeneral). As a result, `printf` doesn't exist. You can force it to exist by including `legacy_stdio_definitions.lib`, but you must include this manually on _every single project_ on windows that uses `printf`, without disrupting linux compilations, like so:
```
terralib.saveobj("benchmark_fannkuchredux", { main = main }, (jit.os == "Windows" and {"\\legacy_stdio_definitions.lib"} or nil) )
```
The larger issue here is that terra cannot actually compile any inline-only function because the function immediately ceases to exist. This also applies to things like compiler intrinsics that don't correspond to an actual function symbol. 

The inline-only problem can be fixed by using `llvm.used`, as was done in #405, but this does not fix the stdio specific symbol resolution problem.

We could also consider trying to just manually include the `stdio_legacy_definitions.lib` on all windows compilations (plus our own library, if necessary) and therefore ensure that all of the _standard library functions_ have symbols, at the very least. Whatever option is chosen, using the C standard library's `printf()` function should not result in random, confusing compiler errors that are incredibly hard to track down.

blackhole12:
This fixes #401 by compensating for what is probably an LLVM bug in `copyAttributesFrom` which doesn't actually copy over the `comdat` attributes. Either that, or LLVM doesn't consider `comdat` a traditional attribute. The actual fix here is a single line of code: `newfn->setComdat(fn->getComdat());`. Everything else is either just to improve MSVC compatibility, or is the stdio workaround being removed from the tests.

A separate bug may need to be filed upstream to LLVM.

### DIFF
--- a/src/tcwrapper.cpp
+++ b/src/tcwrapper.cpp
@@ -961,9 +961,10 @@ static void optimizemodule(TerraTarget *TT, llvm::Module *M) {
 
     for (llvm::Module::iterator it = M->begin(), end = M->end(); it != end; ++it) {
         llvm::Function *fn = &*it;
-        if (fn->hasAvailableExternallyLinkage() || fn->hasLinkOnceODRLinkage()) {
+        if (fn->hasAvailableExternallyLinkage() ||
+            fn->getLinkage() == llvm::GlobalValue::LinkOnceODRLinkage) {
             fn->setLinkage(llvm::GlobalValue::WeakODRLinkage);
-        } else if (fn->hasLinkOnceLinkage()) {
+        } else if (fn->getLinkage() == llvm::GlobalValue::LinkOnceAnyLinkage) {
             fn->setLinkage(llvm::GlobalValue::WeakAnyLinkage);
         }
 #if LLVM_VERSION >= 35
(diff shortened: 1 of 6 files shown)


src/tcwrapper.cpp
```diff
@@ -961,12 +961,12 @@ static void optimizemodule(TerraTarget *TT, llvm::Module *M) {
```
elliottslaughter: Github won't let me comment on this, but right before this line is one that says `std::vector<llvm::Constant *> usedArray;` which I think should go away now.

elliottslaughter:
Merged, thanks again!

PR #416, opened 2019-10-25
Feature/switchstat

aiverson:
Adding syntax for a switch statement to the language.
The chosen syntax is
 ```
switch expr do
  case expr then
    block
  case expr then
    block
  else
    block
end
```
This includes the addition of two new keyword tokens `switch` and `case`, but avoids the addition of any other keywords that most implementations would use such as `default`.
This syntax is still up for discussion.

There is no case fallthrough by default; adding optional case fallthrough is future development.

### CODE REVIEW

**Comment by elliottslaughter** [comment]
> There is no case fallthrough by default; adding optional case fallthrough is future development.

We should discuss whether we want to have fall-through or not. Because if we decide not to *ever* support fall-through, then users could write code like:

```
var x = 0
switch a do
case 1 then
  x = 1
case 2 then
  x = 2
end  
```

But if we may eventually implement fallthrough then this code will become unsafe in the future, so we will want to either make it an error (for now) to omit `break` or `return` or else add a very loud notice on the documentation that this is not safe to rely on.

We don't actually need the `do` keyword, do we? The syntax could be:

```
switch expr
case constant then
  stats
case constant2 then
  stats2
else
  stats3
end  
```

Without any ambiguity.

**Comment by elliottslaughter** [comment]
Some arguments for/against fallthrough that I can think of off the top of my head:

Pro:

  * Makes the semantics more similar to C, so they will generally be more familiar.
  * In the case where you actually need it, it is arguably nicer to deal with than say, `goto`.

Cons:

  * Arguably this is a huge footgun in C and all the languages that inherit C's decision.
  * Not using fallthrough is likely to be what users need 90% of the time, this makes that case both more error-prone (if you forget `break`/`return`) and also longer to type (an extra line and half a dozen extra keystrokes).
  * Makes the variable scoping result much more complicated and not-Lua-like.
  * If you absolutely need it, you can replicate fallthrough using `goto`, though whether that's actually better or not is open to question.
  * For the limited case where you really just want multiple cases with the same code block, you could implement a syntax like `case a case b case c then` (i.e. omit the intervening `then` to merge a bunch of case blocks). The main risk of this syntax is if you accidentally add an extra `then` you'll get an empty code block. But this might be better than the alternative of fully-general fallthrough.

**Comment by aiverson** [comment]
I don't think we should ever have default fallthrough, but optional fallthrough seems like a better idea. I like the option you propose with consolidating multiple cases by only using one then, but it doesn't adequately handle things like Duff's device, guards, partial computation resumption, etc. which need fallthrough from one block to another. Using a keyword to explicitly fall through rather than to explicitly avoid fallthrough seems like a better idea in that case, perhaps `continue`.

**Comment by aiverson** [comment]
Regarding the not needing the `do` keyword, we technically don't need the `then` keyword for if statements or this either, or for the `do` keyword in either while or for loops. The question is whether it makes the code more legible, and whether the legibility is worth the concision and/or complexity.

**Comment by elliottslaughter** [comment]
Sorry for being slow getting to this.

The code itself looks fine to me. Probably due to the limitations that exist at the moment we'll want to mark this as experimental so people don't expect it to be too full-featured.

I think the thing that is bugging me about the syntax is the use of two keywords in a row. I.e. you are always forced to write `do case` with no intervening tokens. I can't think of any other Lua constructs that work like this. So going to `switch ... case` would preserve the property that you always have one keyword between each set of expressions/statements.

I would still like to get @zdevito to look at the syntax if possible since this is a language change. I may write up an explicit syntax RFC so that we have that explicitly all in one place.

**Comment by elliottslaughter** [comment]
Added an RFC here: #417. Please review it and see if it makes sense.

**Comment by aiverson** [comment]
Given the discussion in the RFC, is this going to be accepted or not?

**Comment by aiverson** [comment]
@elliottslaughter could you take a look at this?
It looks like it's ready to merge to me, but one of the mac builds is broken for some reason.

**Comment by elliottslaughter** [comment]
Thanks, I'll try to take a look this week.

To be honest the mac failure doesn't make any sense to me. It looks like a path issue (`install/bin/bin/terra` is clearly wrong), but master seems to be fine, and I don't see anything in the PR that has any influence on this. But I don't think this should be a blocker, you can see in the test that failed the `ctest` portion of the test passed immediately before all the issues.

File: src/terralib.lua [hunk 1]
```diff
@@ -3065,6 +3070,19 @@ function typecheck(topexp,luaenv,simultaneousdefinitions)
         local body = checkblock(s.body)
         return copyobject(s,{condition = e, body = body})
     end
+    local function checkexpintegral(re)
+        local e = checkexp(re)
+        if not e.type:isintegral() then
```
elliottslaughter, line 3075: Do we need to check that all the case expressions are (or coerce to?) the same type?

src/terralib.lua
```diff
@@ -3195,6 +3213,18 @@ function typecheck(topexp,luaenv,simultaneousdefinitions)
                 local br = s.branches:map(checkcondbranch)
                 local els = (s.orelse and checkblock(s.orelse))
                 return s:copy{ branches = br, orelse = els }
+            elseif s:is "switchstat" then
+                local cond = checkexpintegral(s.condition)
+                --local br = s.cases:map(checkcasebranch)
+                --local br = terralib.newlist()
+                --local stack = terralib.newlist{s.cases}
```
elliottslaughter, line 3220: Can we remove these comments if they're not needed?

tests/switch.t
```diff
@@ -0,0 +1,144 @@
...
+  test.eq(fn(4), d)
+end
+
+terra foo(a: int): int
+  switch a do
+    case 1 then
+      return 2
+    case 2 then
+      return 5
+    end
```
elliottslaughter, line 16: Just to check, is this end optional? I can see why we allow it (for consistency with spliced quotes), but it seems nicer to be able to elide it:

```
switch a do
  case 1 then
    return 2
  case 2 then
    return 5
  else
    return 3
  end
end
```

tests/switch.t
```diff
@@ -0,0 +1,144 @@
...
+  end
+  return -1
+end
+
+dotest(single, 2, -1, -1, -1)
+
+terra justelse(a: int): int
+  switch a do
+    else
+      return 3
```
elliottslaughter, line 46: Following on the above, should we require `else` to have its own `end`? (For consistency with when we metaprogram the `else`.)

tests/switch.t
```diff
@@ -0,0 +1,144 @@
...
+  end
+  return -1
+end
+
+dotest(justelse, 3, 3, 3, 3)
+
+terra values(a: int): int
+  switch a + (a + a) do
+    case 1 + (1 + 1) then
```
elliottslaughter, line 55: What's going on here? Do we do constant folding, or is this somehow implicitly escaping a Lua expression?

src/terralib.lua
ErikMcClure: LLVM doesn't care if the case expression is the exact same type, only that it resolves to some integral constant it can use to construct the jump statement. If we choose to expand the switch statement to accept other types, it might be worth putting in a more strict requirement, but since any coercion requirement should still allow the implicit conversion of integers to any comparable integer type I don't think it's necessary to implement it right now.

tests/switch.t
ErikMcClure: LLVM does it's own constant folding. The only requirement we have is that whatever LLVM IR is emitted can be folded into a integer constant from the perspective of LLVM. Hence, when this compiles, LLVM sees the constants, folds them itself, and it ends up working.

tests/switch.t
ErikMcClure: The end is currently mandatory, but I do not remember why - @aiverson might remember. We were running into parsing issues with how terra handles blocks which greatly limited our ability to make parts of the syntax optional.

tests/switch.t
ErikMcClure: The `end` here is intended to conclude the entire case/case/else sequence, it's not just for cases. Once we allow `else` to be metaprogrammed, it would need an `end` statement when inside a quote, but there's no reason to give it one inside the switch statement.

src/terralib.lua
aiverson: Oh, yeah. Sorry, forgot about those.

tests/switch.t
aiverson: The syntax you gave in the example is incorrect. You have one too many  `end`s.
`else` doesn't match a case, because that would make the syntax not enforce the 0 or 1 default branches constraint and make linearizing the case statements in the presence of metaprogramming be much harder to write. The else clause is part of the switch statement, not the case clauses. `switch expr do caselist else block end`. Case statements terminate at an `end` consuming the keyword, or at a `case` or `else` not consuming the keyword.

Therefore, you should be able to write
 ```terra
switch a do
  case 1 then
    return 2
  case 2 then
    return 5
else
  return 3
end
```
without an issue, though I don't think the tests currently enforce that. Putting an `end` after the final case like
 ```terra
switch a do
  case 1 then
    return 2
  case 2 then
    return 5
  end
else
  return 3
end
```
is closer to the intended syntax. The `end` there is needed in order to be able to put splices containing additional cases at the end of a list of cases in a switch statement, like
```terra
switch a do
  case 1 then
    return 2
  end
  [generate_case_list(stuff)]
else
  return 3
end
```
which might be useful for tuning instruction locality.

tests/switch.t
aiverson: Else is already metaprogrammable by just putting a splice inside the `else` block.

tests/switch.t
elliottslaughter: Ok, I think I understand now. Will reply below, this seems like an important point that I don't want to get lost in the discussion.

elliottslaughter:
Up above there was discussion of the syntax of `else` inside `switch` and whether or not `end` should be optional before `else`, and whether `else` should require its own `end`. I.e., the difference between:

(Currently implemented syntax:)

```
switch value do
  case 1 then
    ...
  case 2 then
    ...
  end
else
  ...
end
```

(Alternative syntax:)

```
switch value do
  case 1 then
    ...
  case 2 then
    ...
  else
    ...
  end
end
```

My understanding of the state of the discussion is the following:

Arguments for current syntax:

  * Syntactically encodes constraint that there can be one `else`. I.e., `else` belongs to a `switch`, not to a `case` block.
  * Easier to implement. I understand this is for two reasons:

     1. Since it's enforced syntactically, we don't need a check at the point where we splice in quotes via metaprogramming. However, if we were to eventually add the ability to metaprogram the `else`, we'd need to add such a check regardless.
     2. Something to do with modifying Terra's parser which I don't fully understand. (@aiverson please correct me if I'm wrong.)

Arguments for alternative syntax:

  * More consistent, all branches exist at the same level.
  * In my opinion at least, prettier.
  * Extends more naturally to metaprogramming since `else` is already at the level of a `case` block; doesn't introduce inconsistency if we later go down this road.

I'd like to understand more about the challenges with issue (2) above. I think issue (1) may be something we'll have to attack eventually eventually (though I realize it still adds some implementation burden now).

Am I missing anything?

aiverson:
The alternative syntax suggested there breaks the ability to metaprogram it. We have the ability to metaprogram cases right now, at all, only because of the choice of syntax 1.

aiverson:
How do you feel about this syntax?
```terra
switch value do
  case 1 then
    ...
  case 3 then
    ...
  else
    ...
end
```

elliottslaughter:
Ok, I guess I'm confused. I don't see how to do metaprogramming in the syntax given in your most recent example, but it seems possible in the one I suggested.

Basically, my proposal is that the `end` after a `case`/`else` is optional only if followed by a `case`/`else`. So all of the following should be valid:

```lua
switch value do
  case 1 then
    ...
  end -- ok, this is optional
  else
    ...
  end -- mandatory, no following case/else
end
```


```lua
switch value do
  case 1 then
    ...
  -- no end; this is ok because the end was optional in front of else
  else
    ...
  end -- mandatory, no following case/else
end
```

Which then seems like it makes it clear how to do metaprogramming:

```lua
switch value do
  case 1 then
    ...
  end
  [metaprogram_any_number_of_cases_here]
  else
    ...
  end
end
```

And hypothetically you could also metaprogram the `else` in the future, but it's ok if we don't support that in the initial implementation.

Basically, `case`/`else` are completely standalone, you can glom them together by skipping `end`, and you can metaprogramming at any point in the `switch` body.

Does that work?

aiverson:
Wish granted. It already works like that, except for the unnecessary end.

aiverson:
`else`s cannot be standalone. That breaks assumptions of other parts of the syntax. Think about how you would metaprogram an else onto an if/elseif chain.

elliottslaughter:
Can we add a test to make sure `end` can be elided between `case` and `else`?

I guess I can live without the double `end`, but I would strongly prefer consistent indentation (so `case` and `else` should be on the same level, either on a line with `switch` or indented). Like I already said if it's not immediately meta-programmable that's fine.

aiverson:
Terra is whitespace insensitive, so indent it however you feel like, until we have a formatting tool to enforce otherwise.

Sure, I can add some test cases.

elliottslaughter:
Sure, but I mean the code style used in the tests and examples, etc. Which will affect how other people write code because they're likely to copy-and-paste or at least follow the same style.

aiverson:
Making it possible to make a quote with a standalone `else` block is a bad idea. The else keyword is used in multiple places, and having standalone else blocks creates a broken assumption of parallelism with `if` blocks. The else is part of a switch statement, much like the else is part of an if statement. It is not a standalone construct and it must go at the end. It is indented to match the keyword that introduces the construct, an if or a switch.

ErikMcClure:
To clarify, the indentation used on the else statement is done *because* it cannot be metaprogrammed and *shouldn't* be metaprogrammed. It can't be moved between case statements because it *has* to appear at the end of the switch statement. Trying to indent it as if it were a `default:` statement from C++ will just confuse people even more when they inevitably try to rearrange it or stick it somewhere other than the end of the switch statement. This indentation makes it obvious this isn't possible.

elliottslaughter:
I get that, and I agree that `else` shouldn't (ever) be metaprogrammed.

But I'm not sure that the ability to metaprogram `case` blocks should cause us to contrort the default coding style (which is effectively what these tests are) so significantly.

Let me put it this way: What percentage of users are even going to use metaprogramming anyway? For those that don't, should they need to care that, in theory, if you were to metaprogram a switch, the way that you do this is to add an `end` to the last case and splice in new cases before the `else`? No, I don't think they should need to care about this at all.

Anyone coming from any language on the face of the planet will expect `case`s to be a flat list with `else` at the end. I think it's worthwhile to choose a default coding style that maintains this illusion. Is there extra complexity when you start to metaprogram? Sure. But if you get it wrong, you'll get an error; that's what the compiler is for.

I don't think it's worth the cognitive overhead to have this distinction be exposed in the default coding style.

Aside from that, the patch looks fine to me and I'd be happy to take it in its current state.

ErikMcClure:
Ignore the metaprogramming part of this entirely - Anyone coming from any other language will expect that "else" can be relocated anywhere in the case list because that's how every other switch statement works. That's what the indentation change is for. It's to make sure people don't think they can stick "else" anywhere in the list.

To be fair, I did have an argument with Open Skies about the indentation expectations of literally everyone else from any other programming language, but the problem is that our "else" construct *isn't* the same as the "default" case label in most other languages and consequently really shouldn't follow the same indentation rules.

I think if there is potential for confusion, we should add comments to the test cases explaining that "else" is not indented because it is must be at the end of the case list and nowhere else. Simply ignoring this problem isn't going to make it any less confusing. Either people will be confused by 1) the indentation or 2) the fact that you can't relocate the else statement.

PR #419, opened 2019-12-03
Add support for LLVM 9

elliottslaughter:
Thanks again to @ProfFan who did most of the hard work.


elliottslaughter:
Linux is clean now.

macOS is failing due to what seems to be Travis not using a new enough Ruby for Homebrew (sigh, Travis).

I haven't tested Windows yet but we need to get LLVM built there first anyway.

elliottslaughter:
I'm merging this, we can sort out Windows later.

PR #426, opened 2020-01-17
Upgrade LuaJIT to 2.1.0-beta3

elliottslaughter:
As noted in https://github.com/terralang/terra/issues/422#issuecomment-574390811, LuaJIT 2.0 is broken in macOS Catalina. Rather than following Homebrew and maintain custom patches for 2.0, just upgrade to 2.1 where the problem does not exist in the first place.


elliottslaughter:
I've confirmed that this is sufficient to get macOS Catalina 10.15 to build and pass all but 8 tests. Further progress is stalled on having a machine to actually debug on, but this might be enough to be worth taking even without having it 100% fixed, since it's clearly an improvement.

macOS 10.13 seems to be consistently failing `leaktest.t` on some configurations (according to Travis). I'll have to see if I can reproduce this locally. Currently this is the main blocker since it would be a regression (though maybe we don't care because it's only the Make build, apparently).

elliottslaughter:
Well, the infamous `leaktest.t` is still failing. I am unable to reproduce locally, though it occurs non-deterministically on Travis. I was hoping to fix it by upgrading LuaJIT to the newest `v2.1` branch on Git, but apparently it's still broken. So I think this is still a blocker at this point, unfortunately. I suppose we could add an official `luajit2.1` branch to the main Terra repo, but I don't want to pull this into `master` if it's causing failures like this.

elliottslaughter:
As I wrote in https://github.com/terralang/terra/issues/404, I think the failures are fundamental to the design and guarantees that the GC provides in LuaJIT. I've now seen `leaktest.t` fail on all OSes (and really, I've probably been ignoring them for a while, I vaguely recall having to turn off various configurations to avoid this in the past).

It was necessary to update past the most recent LuaJIT beta; there were failures even aside from leaktest.

The tests are now clean so I'm going to merge this.

PR #434, opened 2020-04-14
Support for LLVM 10 and 11

elliottslaughter:
This should now be working. There is a list of caveats below, but it's close enough that I'm going to merge and track remaining failures via issues.

Known issues at time of merge:

 * https://github.com/terralang/terra/issues/476 cannot include immintrin.h
 * https://github.com/terralang/terra/issues/477 test failures in LLVM 11 when dynamically linked
 * ~https://github.com/terralang/terra/issues/478 doesn't build on macOS with LLVM >= 10~
 * https://github.com/terralang/terra/issues/479 tests fail on macOS with LLVM >= 10

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
         env:
           CHECK_CLANG_FORMAT: 1
   build:
-    name: Build
+    name: Build (os=${{ matrix.os }}, llvm=${{ matrix.llvm }}, cmake=${{ matrix.cmake }}, cuda=${{ matrix.cuda }}, static=${{ matrix.static }}, slib=${{ matrix.slib }}, lua=${{ matrix.lua }})
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,6 +82,10 @@ jobs:
             cuda: '1'
           - llvm: '8'
             cuda: '1'
+          - llvm: '9'
+            cuda: '1'
+          - llvm: '10'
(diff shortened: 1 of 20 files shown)


elliottslaughter:
The remaining two issues seem to both be in C header includes:

 * `strerror.t`: tries to `include("string.h)` but it can't find `strerror_r`
 * `expvec.t`: hits a boatload of compile errors trying to include `immintrin.h`

elliottslaughter:
I had to give up on macOS due to some sort of odd build error, but it's finally passing (modulo caveats above).

elliottslaughter:
Actually, based on the error message it may be an issue with the LLVM binaries. It seems they now depend on having XCode installed:

```
make[2]: *** No rule to make target `/Applications/Xcode_12.0.0_fb.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/lib/libxml2.tbd', needed by `bin/terra'.  Stop.
```

If that's the case this may go away if I build from source.

elliottslaughter:
It's still failing with the binary I built on my local machine. Next step is to attempt to build on Github Actions itself (under the theory that the software setup should be compatible). It would suck if LLVM binaries are no longer sufficiently generic to be used everywhere, but if it gets CI passing it will at least be better than nothing.

elliottslaughter:
The build is indeed successful when LLVM is built on Github Actions. I'm still getting issues including header files though, and the macOS image seems not to have the 10.15 SDK installed. So I think I'm stuck at this point.

At this point I'm frustrated enough with macOS that I'm going to go forward with the merge and leave the remaining items to be fixed at a future date.

Issue #355, opened 2019-02-21
Find new CI platform


elliottslaughter:
Re: https://news.ycombinator.com/item?id=19218036

Not urgent, but probably not worth sticking around in the Travis community given what's going on. I'm leaning toward Circle CI, but suggestions welcome.

ProfFan:
It appears that Travis is gone?

elliottslaughter:
It should be here now: https://travis-ci.org/terralang/terra/builds

I will update the badges on the README.

elliottslaughter:
This is somewhat more urgent now due to a recent pricing change:

  * https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
  * https://news.ycombinator.com/item?id=24964601

Needing to regularly reapply for credits does not sound like fun to me. Maybe it's time to dust off Github Actions and see if we can switch to it for all of our build needs.

elliottslaughter:
This keeps getting worse:

https://news.ycombinator.com/item?id=25338983

For now we're fine, but it's possible that this could become urgent in a hurry if we run out of credits and are unable to get more.

elliottslaughter:
Apparently I misunderstood. travis-ci.org is being terminated *completely*. In order to continue, you have to migrate your project (manually?) to travis-ci.com.

https://docs.travis-ci.com/user/migrate/open-source-repository-migration/#frequently-asked-questions

One way or another we have to do something by December 31.

elliottslaughter:
I went ahead and merged #436 after finally getting it up to parity with Travis. We now have Github Actions support in master.

There isn't really any reason to remove Travis, but once December 31 passes I don't anticipate going to any effort to revive it.

elliottslaughter:
This is now complete. We moved to Github Actions for Mac and Linux.

PR #436, opened 2020-04-22
Add support for GitHub Actions

elliottslaughter:
The main reason to do this currently is that Github actions supports macOS 10.15 (Travis currently does not). However, this is potentially also a hedge in case we need to do #355.

The good news is that #435 seems to have fixed the Catalina build, so we should be fully operational on all macOS versions now (with the caveats listed in that issue).


elliottslaughter:
Ok, Github Actions should now be up to more-or-less parity with Travis. The set of tests is not 100% identical, partly because it's easier to specify the build matrix as a true matrix in Github Actions (with a list of exclusions for what we don't support). But coverage should be equivalent at this point.

PR #442, opened 2020-05-05
Re-enabled cuda support.

Mx7f:
Moved the new check for cudahome to after where it could be initialized.


elliottslaughter:
Good catch, thanks!

PR #443, opened 2020-05-06
Allow vararg function definitions

ErikMcClure:
It turns out that Terra can actually forward varargs if you allow vararg function definitions, by using LLVM's built-in `va_list` intrinsics. Here is an example of a logging function:

```
local va_start = terralib.intrinsic("llvm.va_start", {&int8} -> {})
local va_end = terralib.intrinsic("llvm.va_end", {&int8} -> {})

terra Log(level : int, f : rawstring, ...) : {}
  if level >= 0 then
    C.printf(Levels[level])
  end
  var vl : C.va_list
  va_start([&int8](&vl))
  C.vprintf(f, vl)
  va_end([&int8](&vl))
  C.printf("\n");
end
```

It may be possible to extract individual arguments out, but I haven't attempted to use LLVM's `va_arg` intrinsic, and Terra might not like it. This change simply forwards the varargs state to the function definition.

### DIFF
--- a/src/terralib.lua
+++ b/src/terralib.lua
@@ -1103,7 +1103,7 @@ function terra.defineobjects(fmt,envfn,...)
                 if not terra.isfunction(v) or v:isdefined() then
                     local typ = terra.types.placeholderfunction
                     if c.tree.returntype then
-                        typ = terra.types.functype(c.tree.parameters:map("type"),c.tree.returntype,false)
+                        typ = terra.types.functype(c.tree.parameters:map("type"),c.tree.returntype,c.tree.is_varargs)
                     end
                     v = T.terrafunction(nil,c.name,typ,c.tree)
                 end
--- a/tests/vararg.t
+++ b/tests/vararg.t
@@ -1,0 +1,48 @@
+C = terralib.includecstring [[
+#include <stdio.h>
+#include <string.h>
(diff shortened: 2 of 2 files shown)


aiverson:
I don't like needing to use va_start and va_end like this is C. Terra is a higher level language, it should be able to use `...` to access varargs.

Having a function in terralib that turns varargs into an iterable list for consuming in a for loop would be nice, and we should be able to just use `...` to forward varargs to another function.

elliottslaughter:
@aiverson I think these are good points and a starting point for future discussion on what a truly "Terra native" varargs support would look like. On the other hand, being able to support C ABI compatibility is also useful, so I think this patch is worthwhile.

@ErikMcClure could you add a test like what you have in the issue description? I want to make sure we actually test this behavior, if we want to support it.

ErikMcClure:
@elliottslaughter I've added a test that uses a C function to iterate through and validate each example argument using `va_arg`. This only tests passing through the `va_list` parameter.

elliottslaughter:
Looks like llvm.org must have been down, but I plan to merge this. (Will try to restart the tests first, but if it's still down I'll just merge.)

elliottslaughter:
Ok, whatever. llvm.org is still down. I'm just going to merge this.

PR #452, opened 2020-07-06
Add support for Moonjit

elliottslaughter:
This is definitely not ready yet, putting this here mainly so that other people can come along and help if they want to.

This is an initial patch to try out [Moonjit](https://github.com/moonjit/moonjit) (a LuaJIT fork) as a replacement for LuaJIT for Terra. The main motivation is that Moonjit has PPC64le support.

Currently this breaks at the stage where we try to compile the internal Lua files to bytecode:

```
cd build
cmake ..
make VERBOSE=1
```

Fails with:

```
cd /home/eslaught/terra_moonjit/build/src && ../moonjit/bin/moonjit-2.3.0-dev -bg /home/eslaught/terra_moonjit/src/terralib.lua /home/eslaught/terra_moonjit/build/terralib.bc
../moonjit/bin/moonjit-2.3.0-dev: unknown luaJIT command or jit.* modules not installed
```

If you're in the build directory and you want to try that directly, you can do:

```
moonjit/bin/moonjit-2.3.0-dev -bg ../src/terralib.lua terralib.bc
```

Oddly if I start Moonjit manually it reports the JIT is on, so I'm not sure what's going on:

```
$ moonjit/bin/moonjit-2.3.0-dev
moonjit 2.2.0 -- Copyright (C) 2019-2020 moonjit developers. https://github.com/moonjit/moonjit
JIT: ON SSE3 SSE4.1 fold cse dce fwd dse narrow loop abc sink fuse
```


elliottslaughter:
The suggestion from https://github.com/moonjit/moonjit/issues/128#issuecomment-654560074 fixed the JIT module issue.

At this point I can pass 100% of tests on x86 and 98% of tests on PPC64le. There are 13 failures, which I haven't looked at yet:

```
98% tests passed, 13 tests failed out of 536

Total Test time (real) =  11.49 sec

The following tests FAILED:
	 52 - canon2.t (Failed)
	 55 - cconv.t (SEGFAULT)
	 79 - constanttypes.t (Failed)
	 83 - coverage3.t (Failed)
	112 - defer.t (ILLEGAL)
	113 - deferbreak.t (ILLEGAL)
	138 - expvec.t (Failed)
	169 - gettype.t (Failed)
	210 - luabridge.t (Failed)
	211 - luabridge2.t (Failed)
	238 - multiterra.t (SEGFAULT)
	345 - ssimple.t (Failed)
	409 - vecobj.t (Failed)
Errors while running CTest
```

elliottslaughter:
Ok, Regent on PPC64le now passes all tests except Cholesky, which is probably due to missing BLAS. That's additional confirmation that Terra is working as intended on PPC64le.

I intend to have some Regent users test out their larger codes on PPC64le. Assuming that checks out, then I will plan to clean up and eventually merge this branch. I will preserve the ability to build against plain LuaJIT, but Moonjit will become the default (at least in CMake, I haven't decided whether to bother updating Make).

If anyone objects to Moonjit becoming the default, speak up now.

elliottslaughter:
I have added a new variable `TERRA_LUA` which can be `luajit` or `moonjit` (default: `moonjit`), which controls what LuaJIT implementation you build against. For the moment, I plan to support both. Both are tested in CI. I'm making Moonjit the default because there are platforms (like PPC64le) that are not supported at all on LuaJIT, and those users will need this to run.

I asked a couple of our larger Regent users to test with Moonjit on PPC64le, and everything has checked out, so I'm fairly confident this should be a drop-in replacement for the vast majority of users.

PR #457, opened 2020-08-20
Test debug symbols

elliottslaughter:
This PR is intended to introduce two changes:

1. (DONE) Run all tests with and without the debug flag, so we can be sure it doesn't break existing tests.
2. (TODO) Actually test that debug symbols are working (in `gdb` or similar).


elliottslaughter:
Added a test in #733.

PR #459, opened 2020-09-30
support newer binutils

pixelb:
Pass --verbose as well as --trace to the linker,
as without -verbose, newer versions of gold (2.32 at least),
will not output the component object files of archives,
thus causing the lua script that processes the output to fail.


elliottslaughter:
Is this flag supported by macOS? The macOS non-CMake builds seem to be unhappy. (Plus one unrelated failure, but that looks like a CI glitch.)

elliottslaughter:
We switched to CMake a long time ago and not directly query LLVM's CMake configuration files.

PR #461, opened 2020-10-10
Catch heap corruption on windows

ErikMcClure:
Windows [ignores the unhandled exception filter](https://peteronprogramming.wordpress.com/2017/07/30/crashes-you-cant-handle-easily-3-status_heap_corruption-on-windows/) for heap corruption, which results in `terra.exe` simply silently crashing if heap corruption is encountered. This adds a super special exception handler to actually catch the heap corruption and grab the stack trace (which unfortunately isn't very useful but we can at least tell the user what happened instead of silently crashing).


elliottslaughter:
Looks like all tests pass except formatting (did you use a tab? trailing whitespace?).

Will merge when passing. Thanks.

PR #469, opened 2020-11-30
Support for SPIRV Target

lizdulac:
Work in progress: add clang support for SPIRV target... needs some cleaning.


elliottslaughter:
LLVM finally added SPIR-V upstream, so we ended up not needing to do this. Thanks again for working on it!

PR #475, opened 2021-01-13
Formally deprecate Travis support

elliottslaughter:
This completes #355. We now use Github Actions for the CI. Travis support is removed and the badge now displays Github Actions status.


elliottslaughter:
LLVM's download site seems to be down, but I don't think anything in this change should be a problem, so I'll take it anyway.

PR #480, opened 2021-02-03
Additional fixes for LLVM 11.x (11.1) support 

pmccormick:
Looks like you now need to use explicit LLVM string to std::string conversions.  I tested this with both LLVM 10.x and 11.x and all tests passed.  Did not check older versions of LLVM.

Building terra w/ GCC 10.2 failed due to the patches lines.


elliottslaughter:
Looks good, will merge when tests pass. Thanks!

elliottslaughter:
Looks like clang-format is unhappy with the indentation but otherwise things look fine.

PR #495, opened 2021-05-03
Add CMake-based Nix derivation with CUDA support

seylerius:
This PR cleans up a few minor things along the way to creating a functioning Nix derivation for Terra. I switched to a `FindLuaJIT.cmake` module to support system-provided LuaJIT packages. Also included is a workaround for broken CPU detection in LLVM on some systems, resulting in lots of failures when Terra is compiling due to the CPU being reported as `generic`, as well as a minor change to get the system-specified C/C++ compiler from the standard `CXX` environment variable, rather than just assuming `gcc` will be available.

It builds, with all tests passing, on NixOS x86-64.


elliottslaughter:
Hi there,

Thanks for submitting this!

I'd like to take this a bit more piecemeal. That doesn't necessarily have to involve separate PRs if you don't want, but I may directly cherry-pick changes/commits out of your PR to apply them directly, so we can split up the changes that are being made here.

To start with, you might be on a slightly old branch because I see there are merge conflicts in `tcompiler.cpp`.

I'm happy to have some integration with finding a system LuaJIT. My main issue is that I don't want this to be our entire story, because I don't want users to have to install that dependency manually. I haven't read deeply enough into your CMake changes to see if you actually integrate with `GetLuaJIT` or not. If so then disregard this.

I'll also leave a few comments inline.

src/CMakeLists.txt
```diff
@@ -386,6 +388,10 @@ foreach(TERRA_TEST ${TERRA_TESTS})
     COMMAND $<TARGET_FILE:TerraExecutable> ${TERRA_TEST}
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/tests
   )
+  set_tests_properties(
+    ${TERRA_TEST} PROPERTIES
+    ENVIRONMENT "LUA_PATH=./?.lua"
```
elliottslaughter, line 393: I wonder why this doesn't fail currently.

src/tcompiler.cpp
```diff
@@ -312,6 +312,10 @@ int terra_inittarget(lua_State *L) {
     else
         TT->CPU = llvm::sys::getHostCPUName().str();
 
+    if (TT->CPU == "generic") {
+        TT->CPU = "x86-64";
```
elliottslaughter, line 316: I wonder if we can generalize this? There are systems we run on other than x86, so it would be nice not to hard code this quite as closely. E.g., perhaps we could consult `uname` or `/proc/cpuinfo` or similar (at least on Linux).

src/tcompiler.cpp
```diff
@@ -3186,10 +3190,10 @@ static int terra_disassemble(lua_State *L) {
 static bool FindLinker(terra_State *T, LLVM_PATH_TYPE *linker, const char *target) {
 #ifndef _WIN32
 #if LLVM_VERSION >= 36
-    *linker = *sys::findProgramByName("gcc");
+    *linker = *sys::findProgramByName(getenv("CXX"));
```
elliottslaughter, line 3193: 1. Should this be `CC`? I don't think we actually need to require a C++ compiler at runtime.
2. Either way, we should have a fallback in case `CC`/`CXX` is not defined.

src/CMakeLists.txt
ErikMcClure: That's a really good question isn't it 🙃

src/tcompiler.cpp
ErikMcClure: This could be generalized, but this is mostly just a hack to get around LLVM being broken on certain linux installs. Any attempted generalization would have to eventually have a fallback to "x86-64".

src/tcompiler.cpp
ErikMcClure: This logic is essentially the same [as what CMake does](https://github.com/Kitware/CMake/blob/995be1097c35fd3187d147494156bbfb9da7037f/Modules/CMakeDetermineCXXCompiler.cmake). We can switch it to CC and then copy cmake's list of known compilers.

ErikMcClure:
> I'm happy to have some integration with finding a system LuaJIT. My main issue is that I don't want this to be our entire story, because I don't want users to have to install that dependency manually. I haven't read deeply enough into your CMake changes to see if you actually integrate with `GetLuaJIT` or not. If so then disregard this.

This [falls back to GetLuaJIT](https://github.com/terralang/terra/pull/495/files#diff-13cc4c5527c931690ed0397ed2d55bd3afdc21d71adfedf5ee89b99cb345ccffR18) if it can't find the system LuaJIT.

ErikMcClure:
> I'd like to take this a bit more piecemeal. That doesn't necessarily have to involve separate PRs if you don't want, but I may directly cherry-pick changes/commits out of your PR to apply them directly, so we can split up the changes that are being made here.

I don't personally think this is possible, because **ALL** of these changes are required to get Terra to work on Nix, and none of these changes affect Terra on any other common linux distribution. Thus, it is impossible to test any of these changes in isolation because the only system where any change actually fixes a problem is Nix, and it needs all of them.

If you instead would like them piecemeal for the sole purpose of ensuring they don't break an existing pipeline, that's fine, but please note that as of right now the Terra nixpkg is completely broken until all of these changes are merged.

elliottslaughter:
For the CC/CXX issue specifically, I've actually been maintaining this patch for several years, so I know for a fact that Nix isn't the only system that needs this:

https://github.com/StanfordLegion/terra/commit/f9bb5844d7cdf66e54efb66373d3454ff7ceddee#diff-7e47ff76dd7fbe50e3facbb21d2d88a4856e182fccae42d9b3a786a49178bfad

But I get your point. In general I'm happy to support Nix, and I see that most of these changes are challenging to test individually.

src/tcompiler.cpp
elliottslaughter: Help me understand the nature of this bug. Is it x86 specific? Could someone on ARM or PPC hit it, and if so what are they supposed to do about it if we use this patch?

src/tcompiler.cpp
elliottslaughter: I'd be ok with checking `CXX` if `CC` is unset. At that point, I think if the user has a system unusual enough that falling back to `cc` doesn't work, I'm fine with telling them that they have to set one of those two variables.

src/tcompiler.cpp
seylerius: Erik's looked closer at the mechanics of the bug in LLVM than I have, so I can't say for sure whether it'll come up on ARM or PPC. The issue, as best I understand it, is that we'll have to either reimplement CPU detection ourselves, as a fallback, or we have to make an assumption about what `generic` means, as I've got in this pair of lines.

seylerius:
Okay, this force-push clean up the `tcompiler.cpp` conflict. Minor branch-freshness issue, my bad.

src/tcompiler.cpp
ErikMcClure: Everyone can hit it, because LLVM simply returns "generic" [if anything fails](https://github.com/llvm/llvm-project/blob/d480f968ad8b56d3ee4a6b6df5532d485b0ad01e/llvm/lib/Support/Host.cpp#L1212) in direct conflict with the documentation, which is blatantly wrong. Technically any fix we put in here should actually be a patch to LLVM itself, but even trying to use `/proc/cpuinfo` may not always work for weird linux systems. I bring up "weird linux systems" because those are exactly the ones where LLVM's current detection will fail on.

src/tcompiler.cpp
seylerius: Okay, 192ca54e43eb543fbb6a67c76e1b23432444350f adds a failover here, checking `CC`, then `CXX`, then giving up and demanding `gcc`.

src/tcompiler.cpp
elliottslaughter: Ok, I guess this is fine for now. I agree doing our own CPU auto-detection is undesirable. If we had to do anything, I guess we could add a variable like `TERRA_DEFAULT_ARCH` or similar, but I don't know if this is worth it. We can always add it later if people actually run into this.

src/tcompiler.cpp
elliottslaughter: Looks like line 3077 doesn't actually check `linker_name`?

CMakeLists.txt
```diff
@@ -61,7 +61,11 @@ endif()
 include(GNUInstallDirs)
 include(VersionNumber)
 include(BuildDirs)
-include(GetLuaJIT)
+
+# include(GetLuaJIT)
```
elliottslaughter, line 65: Since it's called through `FindLuaJIT`, can remove the comment.

CMakeLists.txt
```diff
@@ -61,7 +61,11 @@ endif()
...
+find_package(LuaJIT REQUIRED)
+find_program(LUAJIT_EXECUTABLE luajit REQUIRED)
```
elliottslaughter, line 67: Should this be set by `FindLuaJIT`?

elliottslaughter:
This mostly looks good. It needs a pass with `clang-format` (run `./format.sh`), but otherwise looks fine.

Do you want to add a test for nix? Doesn't make a difference to me, but it might help avoid breakage in the future. It could follow a style similar to the Ubuntu test:

https://github.com/terralang/terra/blob/master/docker/Dockerfile.ubuntu

src/tcompiler.cpp
ErikMcClure: Fixed

ErikMcClure:
This pull request now passes all existing tests, but is still broken on certain nix systems for some reason, so we may end up adding a nix test like you suggested.

elliottslaughter:
Please let me know if you're ready to merge this (assuming all CI tests still pass).

ErikMcClure:
We cannot merge this until we have resolved the insane whitespace issue that is happening in the tests.

elliottslaughter:
Now that #496 has been merged, what (if anything) do we still want from this PR?

seylerius:
We're going to redo this on top of https://github.com/terralang/terra/pull/496, using the proper Fundament fork.

PR #496, opened 2021-05-18
Fix linker detection and workaround CPU detection

seylerius:
This is an incremental change, isolating the fix for identifying the linker properly, and the work around for the bug in LLVM's CPU detection.


ErikMcClure:
@elliottslaughter Someone else contributed a nix package that fixed the build without needing the rest of our changes, so we are now able to make incremental pull requests as you originally requested.

elliottslaughter:
Sounds good. Will merge when it passes CI.

PR #497, opened 2021-05-21
Add revised Nix derivation

seylerius:
This is a simple Nix derivation that passes all tests, which are run at the end of the build process. It should continue to pass all Windows tests and CI tests, as nothing has been changed except for adding the `default.nix` and patches. It can be built on any system with Nix installed using `nix-build` in the root of the repository, or with CUDA enabled using `nix-build --arg enableCUDA true`.


elliottslaughter:
Thanks. Looks like it passes CI. Do you want to add some Nix tests too?

disable-luajit-file-download.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/cmake/Modules/GetLuaJIT.cmake b/cmake/Modules/GetLuaJIT.cmake
```
elliottslaughter, line 1: On second thought, would there be any problem with putting the two patch files inside the `nix/` directory? That would make it more clear what they're used for and reduce clutter in the top-level directory.

elliottslaughter:
Let me know when you're ready to merge.

seylerius:
If these pass, we're good. The Nix test is going to run long, because I've got the specific version of Nixpkgs pinned via `niv` until I sort out the rest of the split-derivations issue on jnetod's Nixpkgs update.

seylerius:
Okay, sorted that, this latest can use the latest nixpkgs, and builds much quicker as a result (we're not rebuilding a month-old LLVM for the sake of avoiding the split derivations now).

disable-luajit-file-download.patch
aiverson: We really shouldn't have patch files for our own software stored inside our own repo.
We should just actually fix the problem or add the missing configuration mechanism the patch files are replacing.

seylerius: Oh, huh. I didn't even see this, but wound up stashing those in there, yeah. I do think we need to actually provide the configurations these patches are working around, but I don't think now is the time for it. Probably best done as a cleanup and rework of the whole pile of CMake scripts.

elliottslaughter: I'll take this for now, but we can talk about how to do additional cleanups for the scripts as a next step.

elliottslaughter:
Thanks again, merged!

PR #498, opened 2021-05-31
Fix conversions to logical primitive types

### LINKED ISSUE #460: Non-euclidean Boolean Evaluation
Seriously, what the fuck is going on?

```
> terra a() return [bool](2) == false end
> print(a())
false
> terra a() return [bool](2) end          
> print(a())
false
> terra a() return not [bool](2) end
> print(a())
false
```

Tested using [de24c6](https://github.com/terralang/terra/commit/de24c6c12db7d32e5328ce6f2e5caf3a6995d0ce) on Windows.

LegionMammal978:
This appears to fix #460. In the function without `not`, the argument is converted from `int32` to logical `int8` to `bool`. In the function with `not`, the argument is converted from `int32` to logical `int8`, the `not` operator is applied, and the result is converted from logical `int8` to `bool`; the operator ensures that only 0 or 1 will be converted to `bool`. The incorrect results are caused by primitive values being truncated when converted to smaller logical primitive types, instead of being compared against 0.

In this commit, the change in `emitPrimitiveCast` fixes the `int32` to logical `int8` conversion, so that `[bool](256)` does not equal `false`. Similarly, the change in `ConvertPrimitive` fixes the logical `int8` to `bool` conversion, so that `[bool](2)` does not equal `false`. Both comparison expressions were adapted from the `T_not` conversion on line 1542.

Should I add any new tests for this?

### DIFF
--- a/src/tcompiler.cpp
+++ b/src/tcompiler.cpp
@@ -1044,7 +1044,11 @@ struct CCallingConv {
     PointerType *Ptr(Type *t) { return PointerType::getUnqual(t); }
     Value *ConvertPrimitive(IRBuilder<> *B, Value *src, Type *dstType, bool issigned) {
         if (!dstType->isIntegerTy()) return src;
-        return B->CreateIntCast(src, dstType, issigned);
+        if (dstType == Type::getInt1Ty(*CU->TT->ctx)) {
+            return B->CreateICmpNE(src, ConstantInt::get(src->getType(), 0));
+        } else {
+            return B->CreateIntCast(src, dstType, issigned);
+        }
     }
     void EmitEntry(IRBuilder<> *B, Obj *ftype, Function *func,
                    std::vector<Value *> *variables) {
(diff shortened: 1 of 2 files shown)


elliottslaughter:
Thanks for submitting this!

Yes, a test that covers the new behavior would be nice.

elliottslaughter:
Looks like the CI is complaining about code format. The diff in the test should show you how to change it, it you can install clang-format and run ./format.sh

LegionMammal978:
I've added the new test, and I've checked that it fails without this patch and passes with it.

elliottslaughter:
Unfortunately clang-format is not compatible between major versions. We chose to standardize on version 9 to avoid thrashing the code format, but if it's a trouble to switch I can handle it on my end (later today since I'm not at a computer at the moment).

LegionMammal978:
Okay, it looks like I managed to fix the format thrashing.

elliottslaughter:
Thanks again, merged.

PR #507, opened 2021-07-26
Add script to generate dockerfiles for testing

Jcd1230:
Made this while working on #503, figured I would put it up here in case it could be helpful for anyone else.
Just a lua script that creates a bunch of dockerfiles for building Terra in an Ubuntu container for each of the supported LLVM versions (+12), with either cmake or make.

It excludes a couple configurations as per the GitHub CI config.

### DIFF
--- a/docker/.gitignore
+++ b/docker/.gitignore
@@ -1,0 +1,1 @@
+Dockerfile-*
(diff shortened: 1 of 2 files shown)


elliottslaughter:
Thanks, looks useful.

PR #510, opened 2021-09-16
Check download status and fail build if download fails

elliottslaughter:
This commit adds error handling for the case where the LuaJIT download fails in the CMake build.


elliottslaughter:
@seylerius @ErikMcClure @aiverson Can you advise on the best way to fix Nix for this PR? We currently don't do error code checking at all on the LuaJIT download, but the additional code means we probably need a real patch in Nix (or a proper switch to shut the download off).

PR #525, opened 2022-03-31
Support for AtomicRMW instruction

elliottslaughter:
This is needed for atomics on AMDGPU, but it should also fix a long-standing deficiency with CUDA where we haven't able to use proper atomics since we upgraded from NVVM to NVPTX in LLVM 3.8.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -690,6 +690,71 @@ Create a new macro. The function will be invoked at compile time for each call i
 
 True if `t` is a macro.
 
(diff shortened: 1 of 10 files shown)


elliottslaughter:
Seems to be passing, but should probably add more fail tests to make sure the syntax checks are working.

elliottslaughter:
Some other things to check:

* Return value?
* More precise type checking (or just let it go through to LLVM)?
* Documentation

PR #531, opened 2022-04-09
Add support for fast-math optimizations

### LINKED ISSUE #530: Support fast math flags
See https://github.com/terralang/terra/issues/529#issuecomment-1089464282 for why this matters. This is the last major blocker (as far as I can tell) on #471.

LLVM fast math flags are documented [here](https://llvm.org/docs/LangRef.html#fast-math-flags). The main question is how to expose them.

Two options that come to mind are adding a function-level attribute exposed by a setter. E.g.:

```
my_func:setfastmath(true) -- set all flags
my_func:setfastmath("contract") -- set one flag
my_func:setfastmath({"contract", "nnan"}) -- set two flags
```

This is potentially obnoxious because you need to remember to set it on any functions you care about.

A more whole-program approach would be to set this through `terralib.saveobj`, probably hijacking the `optimize` parameter:

```
terralib.saveobj(filename, filetype, functable, arguments, target, {fastmath = true})
```

A final option would be to add a flag like:

`--fast-math <options>`

Which would then be truly whole-program.

elliottslaughter:
Adds support for fast-math optimizations in `terralib.saveobj`.

Fixes #530

TODO:

 * [x] documentation
 * [x] cudalib.toptx
 * [x] choose default fast-math options on GPU? (going to keep optimizations disabled by default, users can enable them)


elliottslaughter:
This is pretty much ready to go. The one remaining question is what default to set for (particularly GPU) fast-math flags.

Originally, Terra used NVVM for CUDA code generation. (We still support this path with LLVM 3.8.) It turns out that NVVM performs the equivalent of LLVM's `contract` fast-math flag by default. In fact, I'm not aware of any way to turn this off. Strictly speaking, this is incorrect, but the cost must be small enough that NVIDIA decided it was worth it. They made the same decision in NVCC, which I guess is not surprising.

With the migration to NVPTX, this goes away. Fast-math flags are encoded in each floating-point LLVM instruction, and NVPTX actually respects these (as opposed to NVVM which applies `contract` whether users asked for it or not). Because the default is to have all flags disabled, this means that users on NVPTX (and therefore, from Terra's perspective, all users on LLVM 5.0 or newer) see a performance regression via NVVM. But it is strictly speaking more correct. In other words, NVVM's choice might introduce some numerical inaccuracy, and NVPTX's default avoids that (at the cost of performance when this is not necessary).

Since it's easy to turn these flags on now, I think it's reasonable to leave them off by default, but again this does result in a performance regression for anyone who has been sticking it out with LLVM 3.8 and doesn't pay attention to the new arguments to `terralib.saveobj` / `cudalib.toptx`.

If anyone has opinions on this, let me know.

elliottslaughter:
CI failure looks like a Nix issue, and my last (nearly identical) commit passed, so I'm going to go ahead and merge this.

PR #537, opened 2022-04-20
Update Docker build and add compatibility check

elliottslaughter:
This makes some minor tweaks to the Docker build and adds a compatibility check. The check verifies that:

  * We only need a minimal set of dependencies to run the Terra test suite (i.e., no LLVM/CMake/Git/etc. in the test container).
  * Terra binaries built on older distros can be run on newer distros (e.g., Ubuntu 18.04 binaries on Ubuntu 20.04).

This change also adds a new "source" variant of the Docker build. It's going to be horrifically slow, but it's the only way I can come up with to deal with the sort of issues reported here: https://github.com/terralang/terra/issues/366#issuecomment-493871533

The change also packages the resulting Terra binary for release.


elliottslaughter:
Need to remember to make this work with CUDA if I want to produce a releasable build.

PR #541, opened 2022-04-22
Initialize CUDA target lazily

### LINKED ISSUE #446: Allow Terra to initialize when LLVM is not build with NVPTX backend
Right now we have a hard dependency on the NVPTX backend of LLVM even when CUDA is not enabled in Terra. This may potentially limit what LLVM binaries we can compile against. If we make the initialization conditional on CUDA support being enabled then we could build with more pre-built LLVM binaries.

Failure looks like:

```
terra/src/terralib.lua:747: failed to initialize target for LLVM Triple: nvptx64-nvidia-cuda (No available targets are compatible with this triple.)
```

elliottslaughter:
This is to avoid dependency on LLVM NVPTX backend when it's not needed (when CUDA code generation is not enabled/used).

Fixes #446.


elliottslaughter:
AppVeyor is throwing a fit, but everything else looks good, so merging.

PR #548, opened 2022-04-28
Add check that geninternalizedfiles.lua finds at least one file

elliottslaughter:
This is a workaround for the inability to check the return code of `io.popen` without Lua 5.2 compat being enabled (which I'm currently not willing to turn on, because it's a potential compatibility hazard).

If `find` (or its Windows equivalent) isn't present on the system, we'll get a list with zero header files in it; rather than generate un-parseable code, error out in `geninternalizedfiles.lua` so that we see the error close to the point where it occurs.

### DIFF
--- a/src/geninternalizedfiles.lua
+++ b/src/geninternalizedfiles.lua
@@ -14,6 +14,13 @@ for i = 2,#arg,2 do
         end
     end
 end
+-- Hack: we can't get the actual return code of the above command
+-- without LUAJIT_ENABLE_LUA52COMPAT (which might break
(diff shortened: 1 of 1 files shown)


elliottslaughter:
For some other options, see https://stackoverflow.com/q/7607384 . However, it's not obvious that any of these options are portable. At a minimum, we'd need to make sure the chosen option (stdout redirect with `os.execute` or exit code capture with `io.popen`) works with all flavors of Unix and Windows. I'm not currently interested in putting time into discovering how to do this (besides the fact that it might be very ugly to code up).

PR #550, opened 2022-04-29
Fix for use of C structs in different target from the defining one

elliottslaughter:
This is a problem that shows up in the AMD GPU backend, but I think it may also fix a related problem in CUDA (and other targets).

Structs imported via `terralib.includec` track the defining target/function, and use this to reconstitute the correct type when used later. This is not correct when generating code for a different target. So we should really only reuse the struct when the target matches.


elliottslaughter:
Passes all Terra and Regent tests with LLVM 11 and CUDA 11.

elliottslaughter:
Confirmed on the original reproducing code on AMD GPU too.

PR #552, opened 2022-05-19
Add support for LLVM 14

elliottslaughter:
At some point we need to come back and seriously think about how to architect the compiler for opaque pointers. But for now, non-opaque pointers aren't gone yet, so we'll just take advantage of that to get by.

Passes on my local machine.


elliottslaughter:
I'm tracking the opaque pointer migration as #553.

PR #559, opened 2022-05-22
Fix build on FreeBSD

remi6397:
This PR fixes build on FreeBSD. Also renamed Makefile to GNUmakefile, since that's what it is and the status quo confused bmake on macOS and (Free)BSD.

PS: Git has hard time diffing moved files properly, so here is the actual patch for sake of completeness:

<details>

```diff
diff --git a/Makefile b/GNUmakefile
index e2de044..e7adb8f 100644
--- a/Makefile
+++ b/GNUmakefile
@@ -7,7 +7,7 @@
 -include Makefile.inc

 # Debian packages name llvm-config with a version number - list them here in preference order
-LLVM_CONFIG ?= $(shell which llvm-config-3.5 llvm-config | head -1)
+LLVM_CONFIG ?= $(shell which llvm-config-3.5 llvm-config14 llvm-config13 llvm-config true | head -1)
 #luajit will be downloaded automatically (it's much smaller than llvm)
 #to override this, set LUAJIT_PREFIX to the home of an already installed luajit
 LUAJIT_PREFIX ?= build
@@ -23,7 +23,7 @@ LLVM_PREFIX = $(shell $(LLVM_CONFIG) --prefix)
 #if clang is not installed in the same prefix as llvm
 #then use the clang in the caller's path
 ifeq ($(wildcard $(LLVM_PREFIX)/bin/clang),)
-CLANG_PREFIX ?= $(dir $(CLANG))..
+CLANG_PREFIX ?= $(dir $(shell $(CLANG) --version |awk '$$1 == "InstalledDir:" { print$$2 }'))
 else
 CLANG_PREFIX ?= $(LLVM_PREFIX)
 endif
@@ -36,6 +36,8 @@ ENABLE_CUDA ?= $(shell test -e $(CUDA_HOME) && echo 1 || echo 0)
 UNAME := $(shell uname)


+PKG_CONFIG = $(shell command -v pkg-config)
+
 AR = ar
 LD = ld
 FLAGS += -Wall -g -fPIC
@@ -149,7 +151,25 @@ ifeq ($(UNAME), FreeBSD)
 SUPPORT_LIBRARY_FLAGS += -lexecinfo -pthread
 endif

-SUPPORT_LIBRARY_FLAGS += -lffi -ledit -lxml2
+LIBFFI_LIBS = $(shell $(PKG_CONFIG) --libs libffi)
+FLAGS += $(shell $(PKG_CONFIG) --cflags libffi)
+ifeq ($(LIBFFI_LIBS),)
+$(error libffi is not installed)
+endif
+
+LIBXML20_LIBS = $(shell $(PKG_CONFIG) --libs libxml-2.0)
+FLAGS += $(shell $(PKG_CONFIG) --cflags libxml-2.0)
+ifeq ($(LIBXML20_LIBS),)
+$(error libxml-2.0 is not installed)
+endif
+
+LIBEDIT_LIBS = $(shell $(PKG_CONFIG) --libs libedit)
+FLAGS += $(shell $(PKG_CONFIG) --cflags libedit)
+ifeq ($(LIBEDIT_LIBS),)
+$(error libedit is not installed)
+endif
+
+SUPPORT_LIBRARY_FLAGS += $(LIBFFI_LIBS) $(LIBXML20_LIBS) $(LIBEDIT_LIBS)

 PACKAGE_DEPS += $(LUAJIT_LIB)
```

</details>


elliottslaughter:
Thanks for working on this!

There seem to be some CI failures. I think `command` may be a shell built-in?

A couple questions/comments:

 1. Did you try running CMake? I'm happy to take patches for Make, but CMake is intended to be our most portable installation method so I'd be curious if it works.
 2. Are you aware of any options for BSD CI? It's a whole lot easier to keep these sorts of things working when it's tested. (We've had BSD patches in the past, for instance.)
 3. Should the new `Makefile` be called `BSDmakefile`? That was what my impression of what the convention was.
 4. You can also get rid of LLVM 3.5 in the `LLVM_CONFIG` search list. We haven't supported LLVM 3.5 for a while.

remi6397:
> There seem to be some CI failures. I think `command` may be a shell built-in?

Yup, fixed.

> 1. Did you try running CMake? I'm happy to take patches for Make, but CMake is intended to be our most portable installation method so I'd be curious if it works.

It didn't work, but I took care of that too.

> 2. Are you aware of any options for BSD CI? It's a whole lot easier to keep these sorts of things working when it's tested. (We've had BSD patches in the past, for instance.)

FreeBSD Wiki has a page on that: https://wiki.freebsd.org/HostedCI

They reference [Cirrus CI](https://cirrus-ci.org/) there.
 
> 3. Should the new `Makefile` be called `BSDmakefile`? That was what my impression of what the convention was.

    If no -f makefile makefile option is given, make will try to open `makefile' then `Makefile' in order to find the specifications.
&mdash; https://www.freebsd.org/cgi/man.cgi?query=make
 
> 4. You can also get rid of LLVM 3.5 in the `LLVM_CONFIG` search list. We haven't supported LLVM 3.5 for a while.

I removed that one and also gave LLVM 13 higher priority than LLVM 14.

elliottslaughter:
@remi6397 Just FYI, I added FreeBSD CI via Cirrus in #561. CMake works. But the GNU Make-based build seems to be having trouble with pkg-config:

```
GNUmakefile:157: *** libffi is not installed.  Stop.
```

https://cirrus-ci.com/task/4772576076496896?logs=main#L17

Any idea what's going on?

If you want to replicate, you can fork my branch here: https://github.com/elliottslaughter/terra/tree/cirrus-ci-gmake

remi6397:
@elliottslaughter I believe that libffi is a separate package that you need to install, [here](https://github.com/elliottslaughter/terra/blob/71bcdd9727a2fcb0e9e82ff77da526cece68a4a6/.cirrus.yml#L16). But from what I see linking against libffi is extraneous, hence not present in CMakeLists.txt.

elliottslaughter:
@remi6397 I think that libffi is already a [dependency of the llvm package](https://www.freebsd.org/cgi/ports.cgi?query=llvm11&stype=all&sektion=all#llvm11-11.0.1_6). And an attempt to add it didn't change the CI failure ([patch](https://github.com/elliottslaughter/terra/commit/767e53e3af216cd530938d1c333a3259988f0e09), [build](https://cirrus-ci.com/build/5352862464081920)).

Unfortunately, we need libffi if LLVM is linked against it ([example failure](https://github.com/elliottslaughter/terra/runs/6581932282?check_suite_focus=true)). We used to try to auto-detect that in the Makefile, but that was a huge mess. (If I recall, `llvm-config` can't be trusted to provide the right answer. But it's been a while and the details are fuzzy.) CMake is different because LLVM provides first-party CMake support, and as long as the package maintainers don't mess it up we can use that to auto-detect everything in a much more sane way (one of the major motivations for moving to CMake in the first place).

For what it's worth, I also tried shutting it off in the FreeBSD build and it seems to [complain about libxml-2.0 next](https://cirrus-ci.com/task/6376809141895168). Which makes me think it's some sort of a pkg-config issue rather than a libffi issue specifically.

Anyway, I'm happy to let this go either way, as I don't have a strong motivation to fix the Makefile. But if you want to take a look the branch is there to play around with.

PR #566, opened 2022-06-12
Multiarch build support via Docker

elliottslaughter:
This PR adds support for building different architectures (currently ARM64 and PPC64le) to Terra. This relies on Docker's BuildKit to support containers with different architectures. Docker in turn emulates the architecture via qemu. This is slow (my estimate is about 3-4&times; slower than native), but that's probably good enough since we don't need comprehensive support.

This PR does **NOT** attempt to improve the actual support for the architectures in Terra. That's a larger issue that will need to be tackled separately. (Both architectures only pass about 96-98% of tests.) All this PR does is add support for the build, so that we know at least that much works, and are able to provide (albeit partially functional) binaries to interested users who are willing to live on the bleeding edge.


elliottslaughter:
Confirmed that on actual POWER9 hardware, the (emulated) PPC64le build passes all but 11 tests:

```
=================
= FAILING tests
=================
vecobj.t
cconv.t
expvec.t
luabridge2.t
ssimple.t
multiterra.t
canon2.t
constanttypes.t
coverage3.t
gettype.t
luabridge.t
=================

536 tests passed. 11 tests failed.
```

This is also my first cross-distro test:

```
$ cat /etc/redhat-release 
Red Hat Enterprise Linux release 8.2 (Ootpa)
```

PR #575, opened 2022-06-25
More fixes for PPC64le calling convention

elliottslaughter:
This PR includes substantial fixes to the PPC64le calling convention for passing structs. In particular, PPC64le now passes the `cconv.t` test, as well as `cconv_more.t` (which is substantially upgraded in this PR).

The test `cconv_more.t` has been upgraded to use metaprogramming, so that it's easier to extend and modify parameters. In particular the test covers:

 * Functions with `N` scalar arguments (where `N` can be as high as 22, though it defaults to 12 for speed in the CI) for types `int8`, `int16`, `int32`, `int64`, `float`, `double` as well as `T[2]` for all the above types.
 * Same, but with a struct of `N` fields as a single argument.
 * Same, but with two arguments of this struct type instead of one.
 * Same as the above three cases, but with various rotations through sets of types designed to test various edge cases.

The test is also set to helpfully print out the failing LLVM bitcode (if any), which makes debugging immensely easier.

TODO:

 - [x] Refactor, particularly to decide where the PPC64le code should go.
 - [ ] Add `cconv_more.t` tests for nested structs (if possible, may not be easy given current framework). Right now we have this through `cconv.t` but it would be good to be more comprehensive.
 - [x] Add `cconv_more.t` tests for arrays (if possible, may not be easy given current framework). Right now we have this through `cconv.t` but it would be good to be more comprehensive.

### DIFF
--- a/src/tcompiler.cpp
+++ b/src/tcompiler.cpp
@@ -1386,7 +1386,11 @@ struct CCallingConv {
                     EmitCallAggReg(B, casted, a->cctype, arguments);
                 } break;
                 case C_ARRAY_REG: {
-                    assert(false && "unimplemented");
+                    Value *scratch = CreateAlloca(B, a->type->type);
+                    unsigned as = scratch->getType()->getPointerAddressSpace();
+                    emitStoreAgg(B, a->type->type, actual, scratch);
+                    Value *casted = B->CreateBitCast(scratch, Ptr(a->cctype, as));
+                    EmitCallAggReg(B, casted, a->cctype, arguments);
                 } break;
                 default: {
                     assert(!"unhandled argument kind");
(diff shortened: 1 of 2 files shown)


elliottslaughter:
With this change I am also currently passing 98.5% of tests:

```
=================
= FAILING tests
=================
benchmark_dgemm.t
constanttypes.t
coverage3.t
dgemm3.t
gemm.t
multiterra.t
sgemm3.t
vecobj.t
=================

540 tests passed. 8 tests failed.
```

elliottslaughter:
I have updated `cconv_more.t` to include various array cases.

This *fails* on x86 with `N` = 10. I'm going to skip this for now and commit the test with `N` = 9, since x86 calling conv isn't in scope for this PR. However, this is something that needs to be looked at later.

But it passes on PPC64le up to `N` = 22.

PR #579, opened 2022-07-07
Fix for type cache initialization on PPC64le

elliottslaughter:
Terra seems to rely on [this code path](https://github.com/terralang/terra/blob/5bb16734ff1ce6afe00cd56be2670dae04531f6a/src/terralib.lua#L4883-L4898) being executed to initialize some type cache machinery. This currently happens only on platforms where debug info is enabled (i.e., x86 only). Therefore on PPC64le, this initialization does not happen, leading to errors down the line.

For now, make sure that initialization is in place so that we don't depend on the user doing `terralib.cast(uint64, 1ULL)` or similar to make their programs work.

### DIFF
--- a/src/terralib.lua
+++ b/src/terralib.lua
@@ -4897,4 +4897,8 @@ function terra.initdebugfns(traceback,backtrace,lookupsymbol,lookupline,disas)
     terra.disas = terra.cast(FP({po,terra.types.uint64,terra.types.uint64},{}),disas)
 end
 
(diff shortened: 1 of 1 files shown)


elliottslaughter:
With this PR, I have the following set of failing tests:

```
=================
= FAILING tests
=================
coverage3.t
multiterra.t
vecobj.t
=================

546 tests passed. 3 tests failed.
```

PR #592, opened 2022-07-27
Upgrade CI to macOS 11

elliottslaughter:
GitHub Actions is removing support for 10.15.

https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/


elliottslaughter:
I need to rebuild LLVM because macOS 11 no longer includes libxml2.

Issue #485, opened 2021-02-23
Switch to ORCv2 code generation


elliottslaughter:
ORCv1 (which we use via the MCJIT wrapper), which has been deprecated for quite a while. At some point it will finally be removed and we'll be forced to switch (e.g., to LLJIT).

I don't have time to do this at the moment, but if someone wants to take a look, we could switch proactively rather than waiting for our build to break before doing it.

ErikMcClure:
I feel obligated to warn you that ORCv2 is what I attempted to use for inNative and it required [several hacks](https://github.com/innative-sdk/llvm-project/commit/21652e54d3e8e0d8f5f5790c9517251db9ce0904#diff-4095a7afc5ee9e936487b669c021d015fd9d7b81be5d362b595adbaaac26882aR921) inside LLVM itself to actually load DLLs properly, and I haven't attempted to upstream those (mostly because those probably aren't the "correct" fix and I just don't want to deal with LLVM right now). While LLVM certainly likes to claim that ORCv2 is what you "should" use, it risks breaking terra on windows in ways that won't be possible to fix until LLVM fixes how it processes symbols on Windows and COFF loading. This is on LLVM 10, which is the latest version supported by terra.

elliottslaughter:
Is the LLVM project at least aware of these issues? It would be nice to make sure they're aware of them so hopefully they're not too eager to remove ORCv1 while they still remain.

ErikMcClure:
I will have to attempt to update my compiler to LLVM 12 and see if they actually fixed ORCv2 on windows, but I don't know when I'm going to have time to do this. I checked the LLVM 12 code on github and there haven't been any significant changes to the areas where we implemented our workarounds, so I suspect the issues are still there. This won't affect Terra's compilation step, but it will potentially make it so Terra cannot link to a DLL correctly with the JIT.

elliottslaughter:
I made a comment in https://github.com/terralang/terra/pull/503#issuecomment-878619532 about the removal of ORCv1 in LLVM 12. Apparently ORCv1 was removed, but MCJIT is still around. Which is frankly bizarre. But apparently we're getting some mileage out of going back to MCJIT, which (assuming it works) may defer the need to think about this a while longer.

elliottslaughter:
We went ahead with the MCJIT path for #503. It would still be a good idea to investigate ORCv2 at some point because presumably MCJIT will get removed eventually, even if it outlived ORCv1.

elliottslaughter:
[...]

PR #596, opened 2022-08-05
Add CI for AArch64 via Cirrus

elliottslaughter:
Does not pass yet, so I won't merge, but this PR can be used to track pass rate for ARM on actual hardware (whereas emulation seems to suffer from additional failures beyond what ARM hardware experiences).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,4 +1,4 @@
-task:
+freebsd_task:
   name: FreeBSD
   freebsd_instance:
     matrix:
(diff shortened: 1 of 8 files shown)


elliottslaughter:
Failure rate drops dramatically with LLVM 11.1.0:

```
99% tests passed, 4 tests failed out of 549
Total Test time (real) =  25.95 sec
The following tests FAILED:
	 65 - class.t (Failed)
	 70 - class6.t (Failed)
	 90 - coverage3.t (Failed)
	417 - vecobj.t (Failed)
```

I suspect we're starting to feel the pressure from not having #485.

elliottslaughter:
All tests are now fixed when running on ARM hardware.

In emulation, I see a segfault in `atomicrmw.t` but otherwise get all passing tests. Note that this is using LLVM 11.1.0 to work around #597.

PR #599, opened 2022-08-17
Update CI configuration for Ubuntu 20.04

elliottslaughter:
Fixes #598.

Jobs that require Ubuntu 18.04 are relegated to Docker.

PR #600, opened 2022-08-17
Move all Linux CI jobs to Docker

elliottslaughter:
Having given up on #599, I am moving all Linux tests to Docker. This is the only sane way to be isolated from periodic changes in the GitHub runners.

Issues to be decided and/or implemented:

 - [x] Makefile on Linux
 - [x] no-static / no-slib
 - [x] luajit/moonjit
 - [x] Review coverage and make sure we're testing what we want/need


elliottslaughter:
Confirmed the binaries on:

  * Linux ARMv8 Ubuntu 18.04
  * Linux x86_64 Ubuntu 20.04 CUDA 11.1
  * Linux PPC64le RHEL 8

PR #602, opened 2022-08-22
Only include formal parameters in function for varargs call

elliottslaughter:
Potential fix for macOS on M1: https://github.com/terralang/terra/issues/597#issuecomment-1221319348.


elliottslaughter:
Nice! Maybe we could split out the remaining three test failures into separate issues? For the `vararg.t`, it might help to get a backtrace (remember to build Terra in debug mode).

PR #614, opened 2022-10-05
Add support for LLVM 15

elliottslaughter:
This builds (to my surprise) but does not pass yet because LLVM is generating some opaque pointers internally that we're not able to deal with right now. I thought initially we might be able to defer the opaque pointer migration, but it looks like we have to (at least minimally).

TODO:
  * [ ] ~Migrate to opaque pointers: https://releases.llvm.org/15.0.0/docs/OpaquePointers.html~
  * [x] `fmin`/`fmax` for `atomicrmw`: https://releases.llvm.org/15.0.0/docs/ReleaseNotes.html#changes-to-the-llvm-ir


elliottslaughter:
I found the switch to turn off opaque pointers. I think I'll take it, though it's really just buying time to defer the transition to LLVM 16. There are some serious architectural issues to deal with, because we have some logic that wants to rely on Clang for generating struct types parsed from C headers.

Passes locally.

PR #617, opened 2023-05-03
Fix missing migration leftover from prev. commit

romariorios:
`options` changed from a pointer (taken as an argument) to a local object, but the arm code still referenced it as a pointer. Most likely the code has been broken on arm ever since but no one checked.


elliottslaughter:
Indeed, we have AArch64 but no ARM32 tests, so probably we just missed this. Will merge once the full CI passes.

PR #624, opened 2023-05-03
Remove call TargetMachine::adjustPassManager which is gone in LLVM 16

elliottslaughter:
Fixes CUDA build in LLVM 16.


elliottslaughter:
Confirmed on Ubuntu 20.04 with CUDA 11.7. Just waiting for the LLVM build to re-upload and then I'll merge.

Issue #553, opened 2022-05-19
Migrate to opaque pointers


elliottslaughter:
Recent LLVM versions have begun migrating to opaque pointers, i.e., pointers with no type. Notably, LLVM 14 added versions of `CreateLoad` etc. that require an explicit type argument (as the element type is no longer specified in the pointer type).

For now, non-opaque pointers aren't actually gone so we can rely on the old code paths. But presumably at some point LLVM will pull the plug and we'll need to migrate.

My initial effort to do the migration didn't go so well because there are a bunch of corner cases. Not all types are clearly labeled in the compiler, and sometimes they seem to be missing entirely (because we're relying on pointers to carry that information). So in order to work in a world with opaque pointers, we need to track all of that and make sure we don't lose it as it's going down the stack.

elliottslaughter:
This is now required as LLVM 15 has removed non-opaque pointers: https://releases.llvm.org/15.0.0/docs/ReleaseNotes.html#changes-to-the-llvm-ir

rongduo:
很高兴收到您的邮件，我会尽快处理的，谢谢。

elliottslaughter:
@rongduo you seem to be using an auto-responder on this issue tracker. Please stop. This is creating spam on the issues. I've seen this several times now and will need to block you if you don't disable the auto-responder.

elliottslaughter:
Actually, it looks like we're going to be forced to straddle this change. According to the official documentation, opaque pointers in LLVM 14 are *not* production ready. So we'll need to keep the non-opaque code paths until we support LLVM 15+ only.

https://llvm.org/docs/OpaquePointers.html#version-support

elliottslaughter:
The first part of this migration went into #639. I think at this point, there is only one nontrivial place in the code where we need pointer types:

https://github.com/terralang/terra/blob/8f43bed2bf8866799bbbe11e6e560f23b66d4e6d/src/tcompiler.cpp#L601-L603

This occurs in the context of trying to get external types out of Clang. Basically, we define a function that includes all of these types as arguments, and then when we need one, we grab that function and fetch the `N`th argument to get its type.

https://github.com/terralang/terra/blob/8f43bed2bf8866799bbbe11e6e560f23b66d4e6d/src/tcompiler.cpp#L598-L600

If you look for `llvm_definingfunction` you'll find it here:

[...]

PR #639, opened 2023-08-23
Modernize use of LLVM pointer types

elliottslaughter:
This is the first part of the migration toward opaque pointers (#553). Starting with the easy cases, I'm trying to reduce the compiler's reliance on looking up pointee types. But this PR may not get 100% of all the cases.

### DIFF
--- a/src/tcompiler.cpp
+++ b/src/tcompiler.cpp
@@ -1381,8 +1381,7 @@ struct CCallingConv {
                                arguments);
             }
         } else {
-            arguments.push_back(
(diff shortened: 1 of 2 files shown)


elliottslaughter:
Ok, I think that's all the easy cases. Everything else either (a) becomes superfluous once pointer types are removed (i.e., there will be no more bitcasts, so it doesn't matter what we bitcast to), or (b) is more challenging (i.e., I need to figure out how to get types safely out of Clang without them being pruned).

Issue #604, opened 2022-08-23
macOS arm64 test failure: cconv_array.t


shoe42:
The `cconv_array.t` test fails on macOS arm64 (M1). macOS 12.5, using system clang and Homebrew LLVM 13. Trimmed output:
```llvm
$ ../build/bin/terra cconv_array.t
running test for int8[0]
running test for int8[1]
running test for int8[2]
running test for int8[3]
running test for int8[4]
running test for int8[5]
running test for int8[6]
running test for int8[7]
running test for int8[8]
running test for int8[9]
cconv_array.t:53: NYI: cannot call this C function (yet)
stack traceback:
	[C]: in function 'caller'
	cconv_array.t:53: in function 'run_test_case'
	cconv_array.t:64: in main chunk
```

elliottslaughter:
I suspect this is a Lua bug.

Can you please help me with this test?

`bug604.c`:

```c
#include <stdint.h>

int8_t f(int8_t x0, int8_t x1, int8_t x2, int8_t x3, int8_t x4, int8_t x5, int8_t x6, int8_t x7, int8_t x8, int8_t x9) {
  return x0 + x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8 + x9;
}
```

`bug604.t`:

```terra
local c = terralib.includecstring [[
#include <stdint.h>

int8_t f(int8_t x0, int8_t x1, int8_t x2, int8_t x3, int8_t x4, int8_t x5, int8_t x6, int8_t x7, int8_t x8, int8_t x9);
]]

terralib.linklibrary("libbug604.dylib")

print(c.f(0, 1, 2, 3, 4, 5, 6, 7, 8, 9))
```

`bug604.lua`:

```lua
local ffi = require("ffi")

ffi.cdef [[
int8_t f(int8_t x0, int8_t x1, int8_t x2, int8_t x3, int8_t x4, int8_t x5, int8_t x6, int8_t x7, int8_t x8, int8_t x9);
]]

local c = ffi.load("libbug604")

print(c.f(0, 1, 2, 3, 4, 5, 6, 7, 8, 9))
```

Then run:

```
clang bug604.c -o libbug604.dylib -shared -Wall
./build/bin/terra bug604.t
./build/bin/terra bug604.lua
```

It should print `45` with each of the last two commands, but I suspect it will fail with both.

elliottslaughter:
Ok, definitely a LuaJIT bug. I'll file an issue shortly. It might help to have:

[...]

PR #642, opened 2023-08-29
Upgrade LuaJIT and restore the full cconv_array.t test on macOS M1

elliottslaughter:
In principle, this should give us a true resolution to #604, by upgrading LuaJIT to pick up https://github.com/LuaJIT/LuaJIT/commit/83954100dba9fc0cf5eeaf122f007df35ec9a604 which addressed https://github.com/LuaJIT/LuaJIT/issues/205.


elliottslaughter:
Confirmed the fix on macOS M2, so assuming CI is happy, I'll merge this.

PR #641, opened 2023-08-25
Enable opaque pointers in LLVM 16

elliottslaughter:
If this PR works, then we'll be past the transition and will be able to complete #553.

PR #643, opened 2023-09-18
Add support for LLVM 17

elliottslaughter:
This rolls up the changes in #641 and then adds more on top to adapt to the removal of the legacy pass manager in LLVM 17. Overall, this seems to be working dramatically better than LLVM 15-16, so I have updated the opaque pointer version bounds to be LLVM 17-only.

Running the test suite locally (macOS x86) produces one test failure:

```
99% tests passed, 1 tests failed out of 550

Total Test time (real) = 143.88 sec

The following tests FAILED:
	 71 - classifyfloatstructs.t (Failed)
Errors while running CTest
```

For once I'm actually done with this ahead of the LLVM 17 final release (above results were with rc4), so I can't merge this until the final release gets cut. But things are looking good overall.

Known issues:

 * Compile time regressions in `compile_time_array.t` and `compile_time_array2.t` forcing those tests to be shut off in LLVM 17. A similar (but less severe) issue is present in `bug372_perf.t`, but it's minor enough I left that one in. Possibly a result of the new pass manager.
 * `constantinits.t` fails on Linux: #644.
 * I was forced to remove the manual inliner. Figure out how to put it back (or if we need it).
 * Performance has not been verified, so I have no idea if it's anywhere near reasonable or not.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,7 +12,7 @@ freebsd_task:
       # LLVM_VERSION: 14
       LLVM_VERSION: 15
       LLVM_VERSION: 16
-      # LLVM_VERSION: 17 # not available as of 2023-09-21
+      LLVM_VERSION: 17
   install_script: pkg install -y bash coreutils cmake gmake llvm$LLVM_VERSION
   script: |
     export CC=cc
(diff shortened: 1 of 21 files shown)


elliottslaughter:
I'm pretty sure `classifyfloatstructs.t` (the remaining failing test) is failing due to the lack of the manual inliner.

elliottslaughter:
I'm tracking the vectorization issue as https://github.com/llvm/llvm-project/issues/67076

elliottslaughter:
I'm disabling `constantinits.t` for LLVM 17 Linux. Tracking as #644.

PR #652, opened 2023-11-21
Support for SPIR-V code generation

elliottslaughter:
This is a new approach to the work we originally began in #469. Since the last time we put effort into this, Intel integrated a SPIR-V target directly into LLVM itself. The new target was first available in LLVM 15 and has improved dramatically over the last few releases.

Therefore, this puts us on a much better footing than #469 which fundamentally relied on some infrastructure in Clang (because the corresponding LLVM target did not actually exist and it was faked entirely on the Clang side).

I'm currently in the process of verifying that this new SPIR-V support works and is able to generate code that is compatible with, say, OpenCL.

### DIFF
--- a/src/tcompiler.cpp
+++ b/src/tcompiler.cpp
@@ -2828,7 +2828,12 @@ struct FunctionEmitter {
 #if LLVM_VERSION < 170
                         return B->CreateBitCast(v, toT->type);
 #else
-                        return v;
+                        if (fromT->type->getPointerAddressSpace() !=
+                            toT->type->getPointerAddressSpace()) {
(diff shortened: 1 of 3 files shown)


elliottslaughter:
This is at a point where I'm going to take this. SPIR-V code generation output at least works for basic cases. It still requires the LLVM-SPIRV translator to get working SPIR-V, but this is a bug in LLVM's SPIR-V backend, not in Terra. I've fixed some latent bugs in address space handling in Terra so this should have benefits to other backends as well (though I'm not sure anyone will notice).

Issue #657, opened 2024-05-08
Cannot find system headers on macos apple silicon


renehiemstra:
I get an error that the system headers are not found when including them using `terralib.includec(...)`.

I am running on macos Sonoma 14.4.1 using an M3. My C++ compiler is AppleClang 15.0.0.15000309. I can reproduce the issue for different versions of llvm, ranging from 13-17.

commenting out the following lines in terralib.lua solves the issue for me
```
-- Obey the SDKROOT variable on macOS to match Clang behavior.
local sdkroot = os.getenv("SDKROOT")
if sdkroot then
    args:insert("-isysroot")
    args:insert(sdkroot)
end
```

PR #659, opened 2024-05-17
RAII

renehiemstra:
I've closed #658 [(https://github.com/terralang/terra/pull/658)] and reopened it here in order to get a clean git commit history. This second attempt is implemented from scratch.

I'll summarize the pull request and the revised design:

**Objective:** The point of the contribution is to add some metamethods to enable RAII in order to implement smart containers and smart pointers, like `std::string`, `std::vector` and `std::unique_ptr`, `std::shared_ptr`, `boost:offset_ptr` in C++.

The design does not introduce any breaking changes. No new keywords are introduced. Heap resources are acquired and released using the regular C stdlib functions such malloc and free, leaving memory allocation in the hands of the programmer.

**New metamethods:** The following metamethods are implemented:
1. `__init(self : &A)`
2. `__dtor(self : &A)`
3. `__copy(from : &A, to : &B)`

If implemented, these methods are inserted judiciously during the type checking phase, implemented in terralib.lua. All these metamethods can be implemented as macro's or as terra functions.

I will explain the metamethods in some more depth:

A **managed type** is one that implements the above metamethods. In the following I assume `struct A` is a managed type.
- `__init` is used to initialize managed variables:
```
    A.metamethods.__init(self : &A)
```
The implementation checks for an `__init` metamethod in any defvar statement: `var a : A` and applies it if it exists.
- `__dtor` can be used to free heap memory
```
    A.metamethods.__dtor(self : &A)
``` 
The implementation adds a deferred call to `__dtor` (right before any return statements) for any variable local to the current scope that is not returned. Hence, it is tied to the lifetime of the object. `__dtor` is also called before any copy-assignment.
- `__copy` enables specialized copy-assignment and, combined with `__init`, copy construction.  `__copy` takes two arguments, which can be different, as long as one of them is a managed type, e.g.
```
    A.metamethods.__copy(from : &A, to : &B)
```
and / or
```
    A.metamethods.__copy(from : &B, to : &A)
```
`__copy` can be an overloaded function. In object construction, in case of `b` below, `__copy` is combined with `__init` to perform copy construction: 
```
    var a : A
    a.data = ...
    var b = a
```
which is the same as writing
```
    var a : A
    a.data = ...
    var b : A
    b = a
```

I think the above covers all the use cases of smart containers and pointers in c++. I still need to make the metamethods work properly in the case of compositional API's, like  mentioned in #658 (`vector(vector(int))` or a `vector(string)`).

First, let's verify that the CI is passing. Over the next few days I'll add some more tests to test managed data-structures.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /build
+/cmake-build-debug
 /docs/_site
 /docs/_vendor
 /install
@@ -18,6 +19,13 @@
 /tests/dynlib
 /tests/dynlib.exe
 /tests/not_bc
+/tests/class2
+/tests/inline_c.exe
+/tests/objc
+/tests/objc2
+/tests/stdio.exe
+
+/.idea
 
 *.bc
 *.ll
--- a/src/terralib.lua
+++ b/src/terralib.lua
@@ -3692,11 +3692,11 @@ function terra.includecstring(code,cargs,target)
     	args:insert(path)
     end
     -- Obey the SDKROOT variable on macOS to match Clang behavior.
-    local sdkroot = os.getenv("SDKROOT")
-    if sdkroot then
-      args:insert("-isysroot")
-      args:insert(sdkroot)
-    end
+    --local sdkroot = os.getenv("SDKROOT")
+    --if sdkroot then
+    --  args:insert("-isysroot")
+    --  args:insert(sdkroot)
+    --end
     -- Set GNU C version to match value set by Clang: https://github.com/llvm/llvm-project/blob/f77c948d56b09b839262e258af5c6ad701e5b168/clang/lib/Driver/ToolChains/Clang.cpp#L5750-L5753
     if ffi.os ~= "Windows" and terralib.llvm_version >= 100 then
       args:insert("-fgnuc-version=4.2.1")
--- a/src/terralib.lua
+++ b/src/terralib.lua
@@ -2762,6 +2762,29 @@ function typecheck(topexp,luaenv,simultaneousdefinitions)
         return checkcall(anchor, terra.newlist { fnlike }, fnargs, "first", false, location)
     end
 
+    --check if metamethods.__init is implemented
+    local function checkmetainit(anchor, reciever)
+        if reciever.type and reciever.type:isstruct() then
+            if reciever:is "allocvar" then
+                reciever = newobject(anchor,T.var,reciever.name,reciever.symbol):setlvalue(true):withtype(reciever.type)
+            end
+            if reciever.type.metamethods.__init then
+                return checkmethodwithreciever(anchor, true, "__init", reciever, terralib.newlist(), "statement")
+            end
+        end
+    end
+
+    local function checkmetainitializers(anchor, lhs)
+        local stmts = terralib.newlist()
+        for i,e in ipairs(lhs) do
+            local init = checkmetainit(anchor, e)
+            if init then
+                stmts:insert(init)
+            end
+        end
+        return stmts
+    end
+
     local function checkmethod(exp, location)
         local methodname = checklabel(exp.name,true).value
         assert(type(methodname) == "string" or terra.islabel(methodname))
@@ -3309,9 +3332,13 @@ function typecheck(topexp,luaenv,simultaneousdefinitions)
             elseif s:is "defvar" then
                 local rhs = s.hasinit and checkexpressions(s.initializers)
                 local lhs = checkformalparameterlist(s.variables, not s.hasinit)
-                local res = s.hasinit and createassignment(s,lhs,rhs) 
-                            or createstatementlist(s,lhs)
-                return res
+                if s.hasinit then
+                    return createassignment(s,lhs,rhs)
+                else
+                    local res = createstatementlist(s,lhs)
+                    res.statements:insertall(checkmetainitializers(s, lhs))
+                    return res
+                end
             elseif s:is "assignment" then
                 local rhs = checkexpressions(s.rhs)
                 local lhs = checkexpressions(s.lhs,"lexpression")
--- a/src/terralib.lua
+++ b/src/terralib.lua
@@ -3247,9 +3247,59 @@ function typecheck(topexp,luaenv,simultaneousdefinitions)
         return newobject(anchor,T.assignment,lhs,rhs)
     end
 
+    local function checkmetadtors(anchor, stats)
+        --extract the return statement from `stats`, if there is one
+        local function extractreturnstat()
+            local n = #stats
+            if n>0 then
+                local s = stats[n]
+                if s:is "returnstat" then
+                    return s
+                end
+            end
+        end
+        local rstat = extractreturnstat()
+        --extract the returned `var` symbols from a return statement
+        local function extractreturnedsymbols()
+            local ret = {}
+            --loop over expressions in a `letin` return statement
+            for i,v in ipairs(rstat.expression.expressions) do
+                if v:is "var" then
+                    ret[v.name] = v.symbol
+                end
+            end
+            return ret
+        end
+        --check if a __dtor metamethod is implemented for the type corresponding to `sym`
+        local function checkdtor(name,sym)
+            local mt = sym.type.metamethods
+            if mt and mt.__dtor then
+                local reciever = newobject(anchor, T.var, name, sym):setlvalue(true):withtype(sym.type)
+                return checkmethodwithreciever(anchor, true, "__dtor", reciever, terralib.newlist(), "statement")
+            end
+        end
+
+        --get symbols that are returned in case of a return statement
+        local rsyms = rstat and extractreturnedsymbols() or {}
+        --get position at which to add destructor statements
+        local pos = rstat and #stats or #stats+1
+        for name,sym in pairs(env:localenv()) do
+            --if not a return variable ckeck for an implementation of metamethods.__dtor
+            if not rsyms[name] then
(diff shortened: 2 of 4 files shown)


elliottslaughter:
Thanks for submitting this. Looks like CI is still having trouble.

You removed `__move` in this version?

I think it might help to go through some representative code samples for different patterns and show:

 1. What metamethods does the compiler check for? (I guess all of them, all the time? But this wasn't clear.)
 2. Assuming those metamethods exist, what calls get inserted where?

It sounds like the compositional aspect is still unimplemented here.

renehiemstra:
Thanks for considering the pull request.

Regarding the CI. I forgot to uncomment the SDKROOT code for macos (issue #657 ). Also, it seems like the last two months the CI is broken. In particular, the test `fakeasm.t ` segfaults on ubuntu. Maybe, we can give it another try.

I am preparing some code examples (shall I post them here?) In short, the metamethods are checked all the time. The check for `metamethods.__dtor` is done once in `checkblock(...)` (which checks a scoped environment) and `metamethods.(__init, __copy, __dtor)` are checked  in several parts of `checkassignment(...)`.  These checks are cheap, especially if none of the metamethods are implemented.

I will wait with the compositional aspect until we converge to well tested and supported solution.

elliottslaughter:
A CI job from a week ago passed, except for Nix where they recently ripped out LLVM 11 (required currently for our ARM support): https://github.com/terralang/terra/actions/runs/9168683719

Having said that I see a CI job that just ran seems to have failed localenv.t, so maybe there are some non-deterministic failures: https://github.com/terralang/terra/actions/runs/9263364010/job/25481694052

I'll see about getting Nix patched up and hopefully that'll clarify whether we have additional flaky tests or not.

renehiemstra:
Thanks for looking into that.

In my latest commits I have added the compositional aspects, by recursively generating `__init`, `__copy`, and `__dtor` methods. I have implemented this as a separate extension to terralib in the file `lib/terralibext.t`.

I have chosen to make `__init`, `__copy`, and `__dtor` methods rather than metamethods, since there are occasions where the programmer may want to call these methods explicitly in their code.

Rather than merging with the master branch, its probably better if we merge first with a feature branch. Then others can try it out and it remains clear this is experimental.

I am generating documentation now in a separate markdown document. So we can have a technical discussion about the implementation.

elliottslaughter:
If you merge or rebase against `master` hopefully the Nix issue will be resolved. FreeBSD is still unhappy but I'll need to look at that separately. For now it's fine to ignore FreeBSD in this PR.

I agree we should give users some way to test this. Let me think about this more. Mirroring a branch into the main repo is one option; another would be putting it in behind a build or run flag or variable of some sort.

renehiemstra:
I forgot to uncomment the macos systemcode again, which is why the the CI for macos failed. Can we try it one more time?

I've added the file [/lib/raii.md](https://github.com/renehiemstra/terra/blob/raii/lib/raii.md) to document the feature. Maybe we can use that to kickoff further discussions.

renehiemstra:
Undid accidental deletion of a line in the macos systemcode stuff. Restored to original. Now CI should work.

renehiemstra:
Just ran the CI on my [fork](https://github.com/renehiemstra/terra/actions/runs/9438722297). Except for Windows, everything passes. Seems like the CI is not completely deterministic. I'll investigate a bit further.

renehiemstra:
I ran CI on the master branch about a dozen times. 

* *Windows-2022* failed each time due to a segfault (exit code 139) right before running the testsuite. 
* *Docker(ubuntu-18.04, llvm-17.0.5, luajit, static=1, slib=1, cuda=0, prebuilt)* failed once with segfaults in the tests *fakeasm.t* and *localenv.t*

elliottslaughter:
Ok. I guess I need to run the equivalent tests in master to see if I can reproduce the same bugs.

elliottslaughter:
Alternatively (since realistically I will continue to be busy for a while), you could run `master` in your own fork to see if you can get it to reproduce there. Given the odds are around ~10% you'll need to run it a fair number of times (in expectation) to see the failure, if it exists on that branch.

renehiemstra:
I did exactly that. You can see the results [here](https://github.com/renehiemstra/terra/actions/runs/9439670743)

elliottslaughter:
I put in a couple PRs to try to address the `fakeasm.t` failure, which I think are not actually working. (But at least `fakeasm.t` uses `snprintf` now.)

My current thought is to try to build LLVM 17 with Address Sanitizer to see if I can catch anything. I'm not sure whether it will actually work because LuaJIT tends not to play nice with such tools. My initial attempt building Terra only with Address Sanitizer tripped immediately over some LLVM initialization routine, which was not helpful.

elliottslaughter:
I have tried a couple different variations on builds with Address Sanitizer on Linux x86, and have been unable to reproduce any crashes at all. Not sure what that indicates. My Mac builds in #662 also exhibit the same behavior: when I build LLVM with Address Sanitizer, all my crashes go away.

elliottslaughter:
I fixed this in #662 so hopefully those crashes will go away now.

elliottslaughter:
And the FreeBSD build should also be fixed now.

renehiemstra:
Fantastic! I'll have time to check this next week.

renehiemstra:
Great! CI seems to be [working](https://github.com/renehiemstra/terra/actions/runs/9702674617).

.gitignore
```diff
@@ -1,6 +1,7 @@
 /.idea
 /cmake-build-debug
 /build
+/cmake-build-debug
```
elliottslaughter, line 4: This is now a duplicate.

lib/raii.md
elliottslaughter, line 1: If you put this file under `docs/` then it will appear on the website once this PR is accepted, which I think would be nice.

I'm not sure how much it matters but I could suggest some formatting changes for the document itself:

  * Capitalizing language names consistently: C++ not c++
  * No need italicize language names: C++ not *C++*
  * For code, types and function/method names, use code style not italics: `std::string` not *std::string*
  * Have you run a spell checker?
  * For code blocks, you can generalize use ```` ``` ```` by itself and do not need to also indent
  * When you have a code block inside a list, I'm pretty sure GitHub's markdown is smart enough to make the code block part of the list item if you indent the entire block starting with the ```` ``` ```` line

I know there was quite a bit of effort that went into deciding this design, maybe that's worth documenting here too under a "Design decisions" section or similar?

lib/raii.md
```diff
@@ -0,0 +1,152 @@
...
+* The implementation is not aware of when an actual heap allocation is made and therefore assumes that a managed variable always carries a heap resource. It is up to the programmer to properly initialize pointer variables to nil to avoid calling 'free' on uninitialized pointers.
+* Tuple (copy) assignment (regular or using `__copy`) are prohibited by the compiler in case of managed variables. This is done to prevent memory leaks or unwanted deletions in assignments such as
+```
+    a, b = b, a
+```
+* Currently, there is no way to prevent unwanted calls to `__dtor` in cases such as the following. Consider
+```
+terra foo()
+    var b : A
+    return bar(b)
```
elliottslaughter, line 135: This appears to be pass-by-value so is `b` not copied on being passed to `bar`?

src/terralib.lua
```diff
@@ -555,13 +555,13 @@ function T.terrafunction:setinlined(v)
     assert(self:isdefined(), "attempting to set the inlining state of an undefined function")
     self.definition.alwaysinline = not not v
     assert(not (self.definition.alwaysinline and self.definition.dontoptimize),
-           "setinlined(true) and setoptimized(false) are incompatible")
+            "setinlined(true) and setoptimized(false) are incompatible")
```
elliottslaughter, line 558: I don't begrudge removing the trailing whitespace in most of this file, but this change seems to not follow the rest of the formatting of the code generally.

Is there a standardized code formatter you used to generate this formatting? We should probably talk about this.

I will say that the whitespace formatting changes do make the diffs rather hard to read, simply due to sheer volume. I'm not sure whether they can be easily backed out or applied separately. I suppose one option is to apply the format to the master branch first and rebase/merge this branch on top.

elliottslaughter (commented):
I'll need to get back to this, but I left some initial comments.

Overall in terms of the code, it would be easier to review this if we either (a) removed formatting changes to `terralib.lua` or (b) applied those in a separate PR and then merged this PR on top of that one. I realize both approaches require effort but at the moment it's difficult to review simply due to the volume of whitespace changes we're talking about.

.gitignore
renehiemstra: Removed

lib/raii.md
renehiemstra: Improved documentation and added under docs/raii.md

lib/raii.md
renehiemstra: Yes, but terra performs shallow copies. Indeed, the example would also hold if passed by reference.

Honestly, it's a very bad code pattern that should not compile. I've removed it from the documentation. I'll add a section later on "don'ts"

src/terralib.lua
renehiemstra: I don't know exactly how that happened. I am using CLion for the first time, because it has debug capability for the C++ as well as the Lua side of the compiler stack. Maybe I pushed a wrong button. 

The implementation has gone through quite some addition and shaving. Maybe it's good I close and reopen the pull request such that we get a clean commit history. That also gives me the change to review all the parts again myself and add some additional documentation here and there.

renehiemstra:
reopened in #680

PR #662, opened 2024-06-19
Add support for LLVM 18

elliottslaughter:
Currently passing all but 5 tests on macOS x86:

```
99% tests passed, 5 tests failed out of 550

Total Test time (real) =  52.38 sec

The following tests FAILED:
	 65 - class.t (SEGFAULT)
	 67 - class3.t (SEGFAULT)
	 68 - class4.t (SEGFAULT)
	 84 - constant2.t (SEGFAULT)
	 85 - constantinits.t (SEGFAULT)
```


elliottslaughter:
The crashes all go away when I build LLVM and Terra with Address Sanitizer. I'm not quite sure what that means right now. I'm trying a couple more options to make sure this isn't the result of unrelated build flag changes. Unfortunately it's going to make debugging this a bit more difficult.

elliottslaughter:
The crashing tests were a result of not clearing out a function analysis manager when deleting a function. This also fixes 644.

elliottslaughter:
FreeBSD CI services are still down so I won't attempt to spin up an LLVM 18 test there.

PR #667, opened 2024-06-22
Update to newer FreeBSD image

elliottslaughter:
Image list here: https://cirrus-ci.org/guide/FreeBSD/

13.2 is not available, so switching to 14.0.


elliottslaughter:
Noting here in case I come looking for it later: I had to unset `MAKEFLAGS` to unbreak the FreeBSD build. I think this has no practical consequence as I haven't seen Linux use `MAKEFLAGS` to pass parallel job parameters around between Make instances, but in theory we could be giving up build parallelism as a result of this.

PR #675, opened 2024-10-08
Add AVX512 support

tpkessler:
Many modern CPUs support AVX512 with 512 bit wide vector instructions. Currently, terra only enables AVX extensions (256 bit) for its `vector` type. This PR enables basic AVX512 features if the host CPU supports it.

### Advantages

This potentially speeds up computations with the `vector` type by factor of 2 as twice as many instructions can be handled at the same time.

### Disadvantages

None as the old code path (AVX only) is used if the CPU does not support all required AVX512 instruction sets.

### DIFF
--- a/src/tcompiler.cpp
+++ b/src/tcompiler.cpp
@@ -228,6 +228,25 @@ bool HostHasAVX() {
     return Features["avx"];
 }
 
+bool HostHasAVX512() {
+    StringMap<bool> Features;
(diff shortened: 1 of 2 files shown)


elliottslaughter:
This seems reasonable to me, let's just make sure CI passes. Thanks!

tpkessler:
I forgot to run `clang-format` before the commit. This is now fixed. Can you restart the CI please?

Issue #655, opened 2023-11-21
Known issues in LLVM 17


elliottslaughter:
Since we've got a couple of them, listing all the (new) known issues in LLVM 17:

- [ ] #644
- [ ]  #651
- [ ] #654
- [ ] Need to verify performance, as the manual inliner has been disabled and the entire optimization pipeline is brand new (no issue yet)

It's frankly not obvious if these are worth investigating. The transition to opaque pointers shook looks a bunch of problems in LLVM that were not investigated prior to the release. We may just find that some of these go away with LLVM 18 or subsequent versions.

elliottslaughter:
I don't have an issue for the performance comparison, so pasting the results here for posterity.

Pennant:

```console
$ rm terra
$ ln -s terra-llvm16/build terra
$ ./regent.py examples/pennant.rg -fcuda 1 -ll:gpu 1 /scratch/eslaught/resilience/experiment/pennant.tests/leblanc_long4x1000/leblanc.pnt -npieces 1 -numpcx 1 -numpcy 1 -seq_init 0 -par_init 1 -prune 30 -ll:csize 12000 -ll:fsize 12000 -ll:zsize 12000 -level 5
Elapsed time = 1.121140e+02 (total)
Warning: Skipping sequential validation
INIT TIME =   8.470 s
ELAPSED TIME = 105.843 s

$ rm terra
$ ln -s terra-llvm20/build terra
$ ./regent.py examples/pennant.rg -fcuda 1 -ll:gpu 1 /scratch/eslaught/resilience/experiment/pennant.tests/leblanc_long4x1000/leblanc.pnt -npieces 1 -numpcx 1 -numpcy 1 -seq_init 0 -par_init 1 -prune 30 -ll:csize 12000 -ll:fsize 12000 -ll:zsize 12000 -level 5
Elapsed time = 1.106060e+02 (total)
Warning: Skipping sequential validation
INIT TIME =  10.378 s
ELAPSED TIME = 104.369 s

$ rm terra
$ ln -s terra-llvm20-old/build terra
$ ./regent.py examples/pennant.rg -fcuda 1 -ll:gpu 1 /scratch/eslaught/resilience/experiment/pennant.tests/leblanc_long4x1000/leblanc.pnt -npieces 1 -numpcx 1 -numpcy 1 -seq_init 0 -par_init 1 -prune 30 -ll:csize 12000 -ll:fsize 12000 -ll:zsize 12000 -level 5
[...]

PR #677, opened 2024-11-02
Update release packages

ligurio:
On Ubuntu 22.04 packages `libpolly-18-dev` and `libomp-18-dev` are required for building Terra.

Follows up commit 8c279c38e8a9 ("Update release packages to use LLVM 18.").

### DIFF
--- a/release/share/terra/README.md
+++ b/release/share/terra/README.md
@@ -124,7 +124,7 @@ If the binary releases are not appropriate, then you can also build Terra from s
 On recent versions of Ubuntu, you can get these dependencies with:
 
 ```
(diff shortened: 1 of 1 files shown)


elliottslaughter:
I'm actually waffling on whether I should revert the recommended LLVM to 16 due to #655. LLVM 18 is required for AMD GPU support (all versions) and for NVIDIA H100 GPUs, but there is evidence that on the GPU models supported by both, LLVM <= 16 actually provides better performance.

ligurio:
(Sorry, missed your comment.)

> I'm actually waffling on whether I should revert the recommended LLVM to 16 due to https://github.com/terralang/terra/issues/655. LLVM 18 is required for AMD GPU support (all versions) and for NVIDIA H100 GPUs, but there is evidence that on the GPU models supported by both, LLVM <= 16 actually provides better performance.

I would write these limitations in the README and left choosing LLVM version up to user.

PR #685, opened 2025-06-10
Bump up multiarch distro version to fix build errors

elliottslaughter:
Attempting to fix ongoing build errors in multi-arch builds by bumping the distro version.

Sample failure: https://github.com/terralang/terra/actions/runs/15550662007/job/43780521058

This used to work so presumably this is a QEMU or runner bug.


elliottslaughter:
Ubuntu 20.04 is still hitting this in the middle of the LuaJIT build:

> cc: internal compiler error: Segmentation fault signal terminated program cc1

Surely this must be a QEMU bug. We could try updating to 22.04, but at this point it's probably just luck whether we can find a combination that works.

elliottslaughter:
I gave up on this and went with #687 instead. We're still out PPC64le coverage with this change but at least ARM is back. If QEMU fixes the emulation we could put PPC64le back here.

PR #697, opened 2026-01-30
Support for LLVM 19

dawsonfrakes:
Follow up to #684.

It took two attempts a week apart, but I finally got it. Tested (except for FreeBSD) with CI and passing. The key for newer LLVM versions (that took three hours to figure out) on Linux seems to be using `-Wl,--start-group` and `-Wl,--end-group` for LLVM libraries since apparently there's a cyclic dependency now.

I no longer see the error `llvm: JIT has not been linked in.` either, which might be due to the fix I implemented for Windows #694 since it had a similar issue?

This should also allow for LLVM 20+ work to begin if `terralang/llvm-build` gets updated.

Thanks @elliottslaughter and @KaruroChori for your work in #684!


CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ set(TERRA_ENABLE_CUDA ${CUDA_FOUND} CACHE BOOL "Build Terra with support for CUD
 if(DEFINED TERRA_STATIC_LINK_LLVM)
   set(DEFAULT_TERRA_SLIB_INCLUDE_LLVM ${TERRA_STATIC_LINK_LLVM})
 else()
-  if(WIN32)
+  if(WIN32 OR LLVM_VERSION_MAJOR GREATER 18)
```
elliottslaughter, line 20: To be clear, would you expect `DEFAULT_TERRA_SLIB_INCLUDE_LLVM` to work with LLVM 19? I.e., is it just flipping the default or required?

CMakeLists.txt
```diff
@@ -26,6 +26,13 @@ endif()
 set(TERRA_SLIB_INCLUDE_LLVM ${DEFAULT_TERRA_SLIB_INCLUDE_LLVM} CACHE BOOL "Include LLVM in Terra static libraries")
 set(TERRA_STATIC_LINK_LLVM ON CACHE BOOL "Statically link Terra against LLVM")
 
+if(LLVM_VERSION_MAJOR GREATER 18)
+  set(DEFAULT_TERRA_WHOLE_ARCHIVE_LLVM OFF)
```
elliottslaughter, line 30: I'm also wondering if `DEFAULT_TERRA_WHOLE_ARCHIVE_LLVM` could work with LLVM 19? Is it broken or is there some other reason to want it off here?

elliottslaughter (commented):
Thanks for pushing this the rest of the way through. I really appreciate it.

Just a couple of questions inline on the new defaults.

dawsonfrakes:
> To be clear, would you expect `DEFAULT_TERRA_SLIB_INCLUDE_LLVM` to work with LLVM 19? I.e., is it just flipping the default or required?

Indeed, it is required on UNIX if we want to avoid `CommandLine Error: Option 'o' registered more than once!` in the tests. I now see the error message for Windows so we could probably add the following below it.

```cmake
# CMakeLists.txt:43
if(LLVM_VERSION_MAJOR GREATER 18 AND TERRA_SLIB_INCLUDE_LLVM)
  message(FATAL_ERROR "TERRA_SLIB_INCLUDE_LLVM is not supported on LLVM 19+")
endif()
```

> I'm also wondering if `DEFAULT_TERRA_WHOLE_ARCHIVE_LLVM` could work with LLVM 19? Is it broken or is there some other reason to want it off here?

It appears to be broken with the same reason as above. `Option 'o' registered more than once!` sadly appears in the tests. I'm using [your commit](https://github.com/terralang/terra/commit/6cb9edb39a4766b1877624035fef8aff5d587c8c) as the basis for WHOLE_ARCHIVE and I can't seem to find a way around disabling it outside of finding and fixing the duplicate command registration.

I'm going to spend a bit more time looking into the duplicate command registration and see if we can't clean that up a bit for newer LLVM versions.

elliottslaughter:
Yes, please add an error message like you suggested.

If we later find a fix for this issue we can always merge that as a separate PR later. But otherwise this looks ready to go.

You can also update the README support table to save me the trouble of doing it in a follow-up commit (add a line for LLVM 19 after this one):

https://github.com/terralang/terra/blob/1d0ac6b7ba761092a4d3e579f9e57d10d045dcf5/release/share/terra/README.md?plain=1#L155

elliottslaughter:
Thanks again for figuring this out!

elliottslaughter:
@dawsonfrakes I updated the LLVM 20 branch and put it in #698. There are binaries so the CI can run now.

I had forgotten there were a couple other unresolved issues discussed at the top of https://github.com/terralang/terra/pull/684#issue-2836804074, one of which I discuss more at https://github.com/terralang/terra/pull/684#issuecomment-2641360045. Unfortunately I'm at the end of my time slice at the moment, but if you have time and desire to work on it, you're certainly welcome to do so (but no pressure).

PR #698, opened 2026-01-31
Add support for LLVM 20, try 2 (and LLVM 21)

elliottslaughter:
Based on https://github.com/terralang/terra/pull/697 and https://github.com/terralang/terra/pull/684.

Everything is fixed now on macOS/Linux/FreeBSD. Windows works with LLVM 20 but fails tests with LLVM 21.

Windows required the addition of a [patch to LLVM](https://github.com/terralang/llvm-build/blob/921ec0456f46851f1a3d0a73bbea7d709f91c6a9/patches/remove-clang-abi-from-attrs-clang-trunk.patch) which we'll live with for now. Hopefully at some point we get a proper fix for https://github.com/llvm/llvm-project/issues/163349 and then can remove the patch. See also discussion in https://github.com/elliottslaughter/terra/pull/3

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -14,6 +14,7 @@ freebsd_task:
       LLVM_VERSION: 18
       LLVM_VERSION: 19
       LLVM_VERSION: 20
+      LLVM_VERSION: 21
   install_script: pkg install -y bash coreutils cmake gmake llvm$LLVM_VERSION
   script: |
     export CC=cc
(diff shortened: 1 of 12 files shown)


elliottslaughter:
Thanks again @KaruroChori and @dawsonfrakes for helping us get this far!

Windows tests are broken on LLVM 21, but everything else is working, so I took this PR.

PR #721, opened 2026-08-01
Fix building against Ubuntu's LLVM packages

### LINKED ISSUE #636: Restore CI test for dynamic linking
#634 removed the CI test for dynamic linking, so we need to put it back. I don't think this is a fundamental issue, but it may be one of packaging (I can't find an LLVM package that works correctly). Alternatively, it is possible that mixing dynamic LLVM + static Clang causes problems and we need to switch to full dynamic to get things working again.

elliottslaughter:
Ubuntu's packages for LLVM are screwy in several ways: Ubuntu like to slice up its packages into multiple pieces (e.g., MLIR is in its own package), but then doesn't filter them out of the CMake configuration, which means that you can get libraries that don't exist. The installed CMake configuration also references libraries without calling `find_package` on them (e.g., for libedit).

This PR is an attempt to keep up with the world of Ubuntu LLVM packaging by (a) filtering out LLVM components that do not exist (as long as nothing else depends on them) and (b) properly adding the dependencies for system libraries that LLVM depends on.

Users on our own LLVM binaries should not be impacted.

Fixes #636


znmeb:
I can test this on `resolute` - I only used your binaries because I couldn't get Ubuntu's LLVM 22 to pass `cmake`. I'd prefer the Ubuntu ones because I have another LLVM/CLANG dependent package that I want to put on the same container.

elliottslaughter:
Yes, if you can try this out and see whether it works for you, that would be helpful. The trouble seems to be that different Ubuntu packages are sometimes broken in different ways, and accounting for all of them is something of a chore. The current patch in this branch seems to account for the most common patterns (but apparently not LLVM 18 from Ubuntu 24.04).

znmeb:
> Yes, if you can try this out and see whether it works for you, that would be helpful. The trouble seems to be that different Ubuntu packages are sometimes broken in different ways, and accounting for all of them is something of a chore. The current patch in this branch seems to account for the most common patterns (but apparently not LLVM 18 from Ubuntu 24.04).

I'm hoping to standardize on LLVM 22 given that it's the preferred version for Terra. Incidentally, this is the other package <https://github.com/grame-cncm/faust>.

elliottslaughter:
@znmeb If you want to test https://github.com/terralang/terra/pull/724 instead, I think I'm more satisfied with the tradeoffs made in that PR, which feels less like a bolted-on series of hacks than what we have to do in this one.

znmeb:
Yeah, I'll test it tomorrow after it lands in master. Do Ubuntu and Debian use the same LLVM packaging scheme? I'm probably going to split the Terra project out into a new one independent of R and that one might be Debian instead of Ubuntu.

elliottslaughter:
I have merged https://github.com/terralang/terra/pull/724, please go ahead and test. I'm closing this one.

